Issue #61, opened 2024-08-21
Inbound sender of an IBC transfer should be a Cosmos address


Farhad-Shabani:
Brought up by @soareschen

PR #62, opened 2024-08-21
fix: define `Participant` type to accept any sender/receiver addresses

Farhad-Shabani:
Closes: #61

## Description
A [`Participant`](https://github.com/informalsystems/ibc-starknet/blob/30744b368404c20a3518185eec5d77f892ae8799/cairo-contracts/apps/ics20-transfer/src/types.cairo#L169-L174) type has been defined to represent both sender and receiver parties:

```rust
pub enum Participant {
    Native: ContractAddress
    External: Array<felt252>
}
```

I chose `Array<felt252>` as the address type of the `External` variant to maintain neutrality relative to the counterparty's characteristics. That's possible, in future, we see an IBC connection between Starknet and non-Cosmos chains. 
Also, I opted for an `Array` instead of a single `felt252` because a Cosmos address as bech32 string, when serialized, exceeds the size of a single `felt252`. The same might hold true for addresses in other chains.


Farhad-Shabani:
@soareschen, as we discussed, I've updated the `Participant` type. However, clippy is throwing an unexpected error. Everything works fine on my machine. I can't reproduce locally.

Issue #75, opened 2024-09-10
Implement `recv_packet` as a channel component, exposed by the IBC core contract


Farhad-Shabani:
## Overview
The goal here is to come up with a minimal implementation that incorporates necessary handlers and components including channel and router components into IBC core contract so that it can properly handle receiving packets into Starknet.

## Proposal
The `recv_packet` handler will live in the IBC channel component, which handles interactions with various IBC apps through a minimal IBC router. Right now, it's just ICS-20. This router maps `port_id` to the contract address of the IBC applications. So, when a packet is received, based on the specified destination `port_id` in the packet, the handler will look up the application's contract address and invokes the pertaining callback method. For instance, the `recv_packet` handler will trigger the `on_recv_packet` method in the application contract.

With this setup, each IBC app is deployed and operated as a separate Cairo contract, but they all interact with a central IBC core contract. This makes the IBC core contract the only one with the authority to perform callbacks on the IBC application contracts.

PR #76, opened 2024-09-10
feat: implement channel and router components to handle `MsgRecvPacket`

Farhad-Shabani:
Closes: #75

## Description
In this PR, I've assembled all the key types, components, and handlers to wire the IBC channel layer to the IBC applications via a minimal router. In our case, each application runs as a separate Cairo contract. Some finer details, like properly computing packet commitments and acknowledgments, are still pending and will be addressed in future iterations.


cairo-contracts/core/src/channel/components/handler.cairo
```diff
@@ -0,0 +1,176 @@
...
+            // Check if another relayer already relayed the packet.
+            match chan_end_on_b.ordering {
+                ChannelOrdering::Unordered => {},
+                ChannelOrdering::Ordered => { return; }
+            };
+
+            let router_comp = get_dep_component!(@self, RouterHandler);
+
+            let maybe_app_address = router_comp
+                .get_app_address(msg.packet.port_id_on_b.compute_key());
```
soareschen, line 118: I think the problem here is that the binding between channel and port should be permanent. So if we know the channel ID, we'd also know which port ID it should be. Similarly, the channel ID determines which counterparty channel/port the packet should contain, which should be validated. All these bindings should be contained in `ChannelEnd` already, so we can have a helper function to validate that using `ChannelEnd`.

Farhad-Shabani: >I think the problem here is that the binding between channel and port should be permanent. So if we know the channel ID, we'd also know which port ID it should be. Similarly, the channel ID determines which counterparty channel/port the packet should contain, which should be validated

On the same page with you. But these are already being validated IIUC. Once a receiving packet is opened, we read the respective `ChannelEnd` from the storage. Then, we either explicitly like [here](https://github.com/informalsystems/ibc-starknet/blob/farhad/impl-channel/cairo-contracts/core/src/channel/components/handler.cairo#L70-L74) or implicitly through [verify_membership](https://github.com/informalsystems/ibc-starknet/blob/farhad/impl-channel/cairo-contracts/core/src/channel/components/handler.cairo#L99) perform the checks. By the way, I'll go over the `recv_packet_validate` once more when everything is settled to optimize the implementation around.

> All these bindings should be contained in ChannelEnd already, so we can have a helper function to validate that using ChannelEnd.

I'm not sure I follow. Are you thinking of a different version of the `ChannelEnd` struct? Right now, I’m just keeping things simple and sticking close to how ibc-go and ibc-rs do it. But there's definitely room for improvement once things came together and stabilized a bit.

soareschen: It seems like the validation only checks whether the counterparty channel/port matches, but not whether the port ID matches the one that the channel is bound to.

> Are you thinking of a different version of the `ChannelEnd` struct?

Yeah, it doesn't really need to be the `ChannelEnd` struct itself. I am thinking more like a channel-specific context for handling a packet. For example, something like:

```rust
struct ChannelPacketHandler {
  pub local_channel_id: LocalChannelId,
  pub local_port_id: LocalPortId,
  pub counterparty_channel_id: CounterpartyChannelId,
  pub counterparty_port_id: CounterpartyPortId,
  pub receive_packet_commitment_storage: Map<Sequence, PacketCommitment>,
  // or
  pub storage: ContractState,
  // etc
}
```

The exact layout of the context doesn't really matter. It is just that if we can pack all the information we have available from the start, and then do the handling, it could make the code cleaner.

Farhad-Shabani: Okay. Got it what you mean now.

> but not whether the port ID matches the one that the channel is bound to.

This also gets checked when retrieving the `ChannelEnd` from the store, since it's done using a pair of `(PortId, ChannelId)`.

I’ve cleaned up the code to make it more streamlined and closer to what you had in mind. In terms of structure, I actually tend to keep things consistent with the current IBC specs **wherever possible**, so this `ChannelEnd` struct is modeled after ICS-04 and the provable store format in ICS-24. 
Based on what you mentioned, I think I've hit the mark. If not, just let me know! I’ll dig into it in the next PR when I’ll be consolidating the implementation by adding tests.

soareschen: > This also gets checked when retrieving the ChannelEnd from the store

Ah alright then. Missed that part of the implicit logic.

> I actually tend to keep things consistent with the current IBC specs wherever possible

Ok, let's finish the implementation first, then we will look at if things can be improved.

PR #78, opened 2024-09-11
imp: cairo/scarb structure suggestions

drspacemn:
## Structure:
- Separate out packages from top-level contracts
- Include packages in virtual manifest
- Remove redundant workspace definitions from workspace packages
- Only define shared resources in top-level dependencies
- Explicit tool version via `.tool-versions` file
- Define all script testing via snforge
- More explicit visibility

## Personal Preferences:
- define all modules in lib.cairo instead of breaking out small file imports
- `.workspace = true` > `= { workspace = true }`
- define package versions manually for more granular control
- configure all sub modules under tests with `#[cfg(test)]` except for `mod mocks` in order to keep mocks segregated from business logic but still buildable via snforge

## Recommendations:
- Include mocks for all components and local testing infrastructure for the components via mocks


Farhad-Shabani (approved):
All the points you mentioned sound great! 

Just as for `{ workspace = true }`, I kept it as-is to stay consistent with the other workspaces. Plus, I found it a bit cleaner to catch at a glance.

On the granular versioning side, since we're still early in development, we'll likely bump the version together for a while until things stabilize. But after, your suggestion definitely makes sense. I also got the meta `starknet_ibc` package back in. 

Only remained local component testing. I opened up #79 to tackle it as a separate effort.

Thanks again for the awesome suggestions and the PR! ✨

justtry1512:
OK. I'M DOINT IT NOW.

PR #84, opened 2024-09-18
Protobuf types in cairo

rnbguy:
- protobuf codec in cairo
- cometbft ibc light client protobuf types in cairo


rnbguy:
The protobuf decoding and encoding works fine.

As for next, we should see how to organize the types. The protobuf library lives in `cairo-contracts/protobuf` and cometbft types in `cairo-contracts/cometbft`. Probably some of the cometbft types should be imported from `cairo-contracts/packages` (e.g. ibc `Height`) and move the `ProtoMessage` impls accordingly.

Also, we need to look at the protobuf codec implementation for any additional refinements or refactors.

cairo-contracts/protobuf/src/types/tag.cairo
```diff
@@ -0,0 +1,60 @@
+#[derive(Copy, Debug, Clone, Drop, PartialEq)]
+pub enum WireType {
+    Varint,
+    Fixed64,
+    LengthDelimited,
+    StartGroup,
+    EndGroup,
+    Fixed32,
+    Unknown,
+}
```
Farhad-Shabani, line 10: Is this variant necessary? Might we ever fall into this?

cairo-contracts/protobuf/src/types/message.cairo
```diff
@@ -0,0 +1,140 @@
...
+
+    fn encode<T, +ProtoMessage<T>>(value: @T) -> ByteArray {
+        let mut bytes = "";
+        value.encode_raw(ref bytes);
+        bytes
+    }
+
+    fn from_any<T, +ProtoMessage<T>, +Drop<T>, +Default<T>>(any: @Any) -> T {
+        if any.type_url != @ProtoMessage::<T>::type_url() {
+            panic!("unexpected type URL");
```
Farhad-Shabani, line 36: I agree about the panics. It would be better to use `SyscallResult` or `Option`, where appropriate.

cairo-contracts/protobuf/src/types/message.cairo
```diff
@@ -0,0 +1,140 @@
...
+
+    fn to_any<T, +ProtoMessage<T>>(value: @T) -> Any {
+        Any { type_url: ProtoMessage::<T>::type_url(), value: Self::encode(value) }
+    }
+
+
+    fn decode_field<T, +ProtoMessage<T>, +Drop<T>>(
+        field_number: u8, ref value: T, serialized: @ByteArray, ref index: usize, bound: usize
+    ) {
+        assert(bound <= serialized.len(), 'invalid bound');
```
Farhad-Shabani, line 49: It would be good to define an `errors.cairo` to handle possible error messages from there.

cairo-contracts/cometbft/src/ics23.cairo
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl InnerSpecAsProtoMessage of ProtoMessage<InnerSpec> {
+    fn encode_raw(self: @InnerSpec, ref output: ByteArray) {
+        ProtoCodecImpl::encode_repeated_field(1, self.child_order, ref output);
+        ProtoCodecImpl::encode_field(2, self.child_size, ref output);
+        ProtoCodecImpl::encode_field(3, self.min_prefix_length, ref output);
+        ProtoCodecImpl::encode_field(4, self.max_prefix_length, ref output);
+        ProtoCodecImpl::encode_field(5, self.empty_child, ref output);
+        ProtoCodecImpl::encode_field(6, self.hash, ref output);
```
Farhad-Shabani, line 121: Do you think it makes sense to have a `field_number` field within the `ProtoCodec` struct and expose `init()` and `finalize()` methods by the `ProtoCodecTrait` to eliminate the need for the `field_number` arguments in places like here altogether?

Farhad-Shabani (commented):
I had a quick look over your PR and left some initial comments. I'll take a more detailed look later.

Overall, it looks like an amazing job. much appreciated!

Regarding codebase organization, I’m thinking we should place these two new additions under a separate `cairo-libs` directory (as opposed to `cairo-contracts`, where the existing implementation is for crafting necessary components and combining them as a Cairo contract). It makes sense to have their workspace and versioning segregated as well.

cairo-contracts/protobuf/src/types/tag.cairo
rnbguy: You're right ! I forgot to remove this after correcting a silly mistake on my part.

ed9199c

cairo-contracts/cometbft/src/ics23.cairo
rnbguy: The field numbers are associated with the struct fields - part of the struct definition, which should have been auto-generated. The field numbers don't need to be sequential - only strictly monotonic. 

Prost uses macro for this, where each generated field is tagged with its corresponding protobuf details, e.g. field number, types, repeated or not, optional or not.

rnbguy: Ah I can apply your suggestion for input bytes and index - to reduce arguments in the encode/decode methods.

For decoding,
```cairo
decode_instance.branch(limit);
..
decode_instance.field(5, ref self.empty_child);
...
decode_instance.done();
```

For encoding,
```cairo
...
encode_instance.field(5, self.empty_child)
...
```

cairo-contracts/protobuf/src/types/message.cairo
```diff
@@ -0,0 +1,140 @@
...
+        if index < serialized.len() && index < bound {
+            let tag = ProtobufTagImpl::decode(serialized[index]);
+            if tag.field_number == field_number {
```
soareschen, line 52: I think the protobuf specs allows fields to be encoded in any order, but here we assume that the fields are in ascending order. Also, Protobuf allows the same field tag to be present multiple times, by which the last occurance would win out.

In any case, I suppose the current implementation is sufficient for the first iteration for us to move on. We can figure a more robust decoding implementation later on.

For reference, Prost runs the decoding in a loop, with each field decoder repeatedly called when the tag matches. On the other hand, in Hermes SDK I tried a more organized approach, by first splitting the decode buffer into chunks called `ProtoChunk`, and then put them in a `BTreeMap` with the field tag being the key.

cairo-contracts/protobuf/src/types/message.cairo
```diff
@@ -0,0 +1,140 @@
+use protobuf::types::tag::{WireType, ProtobufTag, ProtobufTagImpl};
+use protobuf::primitives::numeric::NumberAsProtoMessage;
+use protobuf::types::wkt::Any;
+
+pub trait ProtoMessage<T> {
+    fn decode_raw(ref value: T, serialized: @ByteArray, ref index: usize, length: usize);
+    fn encode_raw(self: @T, ref output: ByteArray);
+    fn wire_type() -> WireType;
+    fn type_url() -> ByteArray;
```
soareschen, line 9: I think the `type_url` method should be placed in a separate trait, as not all type implements it.

cairo-contracts/protobuf/src/main.cairo
```diff
@@ -0,0 +1,3 @@
+fn main() -> u32 {
```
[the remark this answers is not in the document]
rnbguy: Ah. Forgot to remove the `main.cairo` :sweat_smile:

cairo-contracts/protobuf/src/types/message.cairo
rnbguy: Good point. I see that prost has two traits `Name` and `Message` with `Name: Message` bound to avoid this.

soareschen:
@rnbguy let's get this merge if it is ready.

soareschen (approved):

rnbguy:
hey @soareschen, I pushed my recent changes. But I am stuck at the `to_any` impls. It indicates that it didn't find any `ProtoName` impls for the types - but they are implemented.

cairo-contracts/Scarb.toml
```diff
@@ -5,6 +5,9 @@ members = [
   "packages/contracts",
   "packages/core",
   "packages/utils",
+  "cairo-libs/protobuf",
+  "cairo-libs/cometbft",
+  "cometbft",
```
Farhad-Shabani, line 10: Can `cairo-libs` be having it own Scarb workspace, not mixed here?

rnbguy:
Fixed the trait issue. cairo probably doesn't have supertraits. The implementation is final :slightly_smiling_face:

cairo-contracts/Scarb.toml
rnbguy: Ah. I tried not to separate the workspace - as we still need to sync the `cairo_test` and `snforge_std` among each other.

rnbguy: But I just noticed, the deps are at `dev-deps` - so it's alright. I will move to a separate workspace at project root.

rnbguy: 89a8fad

rnbguy:
We can remove the `init_branch` and `end_branch` calls from the struct impls. Let me do that.

cairo-libs/packages/protobuf/src/types/message.cairo
```diff
@@ -0,0 +1,169 @@
...
+            }
+            let mut item = Default::<T>::default();
+
+            self.decode_field(field_number, ref item);
+            value.append(item);
+        };
+    }
+
+    fn end_branch(ref self: DecodeContext) {
+        // TODO(rano): pop_back is not impl for Array<T>, this is inefficient
```
rnbguy, line 140: Do we have a stack implementation in Cairo ? It seems, the `Array` is actually a queue implementation.

PR #106, opened 2024-11-06
feat: add endpoint to query latest updated height at or below target height

Farhad-Shabani:
## Context

Requested by @soareschen 
> We need a query method from the client contract for [finding the consensus state height](https://github.com/informalsystems/hermes-sdk/blob/main/crates/chain/chain-components/src/traits/queries/consensus_state_height.rs#L24-L28) before a given target height.
This is needed to determine which trusted height to be used on the Cosmos chain to build the UpdateClient header. In the Cosmos chain implementation, we had to get a list of all consensus heights and then sort and find the trusted height, which is not very efficient. It would be good if there are ways to implement this more efficiently on the chain side.

## Description

This PR introduces the `update_height_before` method to the `IClientQuery` interface, and make the comet client to keep track of the heights at which a client is updated. More context: https://github.com/informalsystems/ibc-starknet/pull/106#discussion_r1833314395

### DIFF
--- a/cairo-contracts/packages/clients/src/cometbft/component.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/component.cairo
@@ -518,17 +518,22 @@ pub mod CometClientComponent {
                 return;
             }
 
-            if update_heights.len() == 255 {
+            let len = update_heights.len();
+
+            if len == 255 {
                 update_heights.pop_front().unwrap();
             }
 
             update_heights.append(update_height);
 
-            let mut update_heights_span = update_heights.span();
+            let new_update_heights = if len.is_non_zero()
+                && update_heights.at(len - 1) > @update_height {
+                MergeSort::sort(update_heights.span())
+            } else {
+                update_heights
+            };
 
-            let sorted_update_heights = MergeSort::sort(update_heights_span);
-
-            self.update_heights.write(client_sequence, sorted_update_heights);
+            self.update_heights.write(client_sequence, new_update_heights);
         }
 
         fn write_client_state(
--- a/cairo-contracts/packages/clients/src/cometbft/component.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/component.cairo
@@ -3,8 +3,8 @@ pub mod CometClientComponent {
     use alexandria_sorting::MergeSort;
     use core::num::traits::Zero;
     use starknet::storage::{
-        StoragePathEntry, Map, StorageMapReadAccess, StorageMapWriteAccess,
-        StoragePointerReadAccess, StoragePointerWriteAccess,
+        Map, StorageMapReadAccess, StorageMapWriteAccess, StoragePointerReadAccess,
+        StoragePointerWriteAccess,
     };
     use starknet::{get_block_timestamp, get_block_number};
     use starknet_ibc_clients::cometbft::{
--- a/cairo-contracts/packages/clients/src/cometbft/component.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/component.cairo
@@ -520,7 +520,7 @@ pub mod CometClientComponent {
 
             let len = update_heights.len();
 
-            if len == 255 {
+            if len == 100 {
                 update_heights.pop_front().unwrap();
             }
 
--- a/cairo-contracts/packages/clients/src/cometbft/component.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/component.cairo
@@ -1,5 +1,6 @@
 #[starknet::component]
 pub mod CometClientComponent {
+    use alexandria_data_structures::array_ext::ArrayTraitExt;
     use alexandria_sorting::MergeSort;
     use core::num::traits::Zero;
     use starknet::storage::{
@@ -512,6 +513,10 @@ pub mod CometClientComponent {
             ref self: ComponentState<TContractState>, client_sequence: u64, update_height: Height
         ) {
             let mut update_heights = self.update_heights.read(client_sequence);
+
+            if update_heights.contains(@update_height) {
+                return;
+            }
 
             if update_heights.len() == 255 {
                 update_heights.pop_front().unwrap();
--- a/cairo-contracts/packages/clients/src/cometbft/component.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/component.cairo
@@ -496,7 +496,7 @@ pub mod CometClientComponent {
         fn read_update_heights(
             self: @ComponentState<TContractState>, client_sequence: u64
         ) -> Array<Height> {
-            self.update_heights.entry(client_sequence).read()
+            self.update_heights.read(client_sequence)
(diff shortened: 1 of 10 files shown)


cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -77,6 +78,30 @@ pub mod CometClientComponent {
...
+            let update_heights = self.read_update_heights(client_sequence);
+
+            let len = update_heights.len();
+
+            assert(len > 0, CometErrors::ZERO_UPDATE_HEIGHTS);
+
+            let mut i = 0;
+
+            while i < len {
+                let height = update_heights.at(i);
```
soareschen, line 93: Interesting, do Cairo not have iterators?

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -415,6 +440,26 @@ pub mod CometClientComponent {
...
         }
 
+        /// Returns the heights at which the client has been updated in
+        /// descending order.
+        fn read_update_heights(
+            self: @ComponentState<TContractState>, client_sequence: u64
+        ) -> Array<Height> {
+            let mut heights: Array<Height> = ArrayTrait::new();
+
+            let entry = self.update_heights.entry(client_sequence);
```
soareschen, line 450: Can't we just return the entry here as `Vec<Height>`?

Farhad-Shabani: Not actually. Getting `Vec<Height>` is possible only if we could call `read()` instead of `entry()`, though this method is not available on `Map` containing a `Vec`. 
The `entry` in fact is of type `StoragePath<Vec<Height>>` which gives access to some methods on `Vec`, but not allowing to return the own `Vec`.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -77,6 +78,30 @@ pub mod CometClientComponent {
             comet_client_state.latest_height
         }
 
+        fn update_height_before(
+            self: @ComponentState<TContractState>, client_sequence: u64, target_height: Height
+        ) -> Height {
+            let update_heights = self.read_update_heights(client_sequence);
```
soareschen, line 84: If we need to rebuild a whole array and then iterate over it, it might be better to just read directly from the storage, and then stop reading once we find the correct height.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -470,6 +515,12 @@ pub mod CometClientComponent {
             self.client_sequence.write(client_sequence);
         }
 
+        fn write_update_height(
+            ref self: ComponentState<TContractState>, client_sequence: u64, update_height: Height
+        ) {
+            self.update_heights.entry(client_sequence).append().write(update_height)
```
soareschen, line 521: This does not order the update heights in ascending or descending order. So if there is an UpdateClient with height earlier than the latest height, it would cause the query to return incorrect result.

cairo-contracts/packages/clients/src/cometbft/component.cairo
soareschen: Something like `filter` or `filter_map` could do that. But a loop is fine for now.

cairo-contracts/packages/clients/src/cometbft/component.cairo
soareschen: Something like `filter` or `filter_map` could do that. But a loop is fine for now.

cairo-contracts/packages/clients/src/cometbft/component.cairo
Farhad-Shabani: Oh, that's absolutely right. I still haven't wired this `write_update_height` up anywhere. Meant to achieve a proper ordering with `Vec` yesterday but it's quite tricky as there is only a single `append` method available on `Vec`. I'll look into it more see what I can do.

cairo-contracts/packages/clients/src/cometbft/component.cairo
Farhad-Shabani: I see. That's unfortunate. There isn't any sort of fliter-like methods available.

cairo-contracts/packages/clients/src/cometbft/component.cairo
Farhad-Shabani: I ended up using the `Array` instead of `Vec` so can utilize a library available for sorting. This required me to implement the `Store` trait manually for `Array<Height>` to enable read and write operations once it’s placed in the `Map`. There is only a minor limitation (at least for now): the `Array` size can’t exceed 255.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -470,6 +508,24 @@ pub mod CometClientComponent {
...
+
+            if update_heights.len() == 255 {
+                update_heights.pop_front().unwrap();
+            }
+
+            update_heights.append(update_height);
+
+            let mut update_heights_span = update_heights.span();
+
+            let sorted_update_heights = MergeSort::sort(update_heights_span);
```
soareschen, line 524: Probably a way to reduce the update cost is to check if the given update height is higher than the latest height, then we can just append and not do any merging.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -470,6 +508,24 @@ pub mod CometClientComponent {
             self.client_sequence.write(client_sequence);
         }
 
+        fn write_update_height(
+            ref self: ComponentState<TContractState>, client_sequence: u64, update_height: Height
+        ) {
+            let mut update_heights = self.update_heights.read(client_sequence);
+
+            if update_heights.len() == 255 {
+                update_heights.pop_front().unwrap();
```
soareschen, line 517: I think it is fine to do this, as it is not good to store all past consensus state heights here. We can even limit it to something like 50~100, as the main purpose for this function is mainly to help resolve recent race conditions between multiple relayers.

cairo-contracts/packages/clients/src/cometbft/component.cairo
Farhad-Shabani: Good point 👌🏻 
[6411a82](https://github.com/informalsystems/ibc-starknet/pull/106/commits/6411a827959194162cae527c76295305663c8d48)

cairo-contracts/packages/clients/src/cometbft/component.cairo
Farhad-Shabani: Alright. Set to 100.
[4c5a1a3](https://github.com/informalsystems/ibc-starknet/pull/106/commits/4c5a1a3324d38af51e7376403b013b66834acd98)

PR #111, opened 2024-11-12
ci: arm64 runner for linux and macos

soareschen:
Using the Ubuntu `aarch64` runners from [Github's public preview](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).


Farhad-Shabani (approved):
Thanks a bunch for investigating around. Just a question:

Couldn't we go with Ubuntu 20.04? Asking because I see it generally the more canonical version that most projects run on, thus I was wondering if we could stick with it for consistency. Otherwise looks good to me.

rnbguy:
> Couldn't we go with Ubuntu 20.04?

I just picked up the latest one. I downgraded them to `ubuntu-22.04` as `ubuntu-20.04-arm` is not available.

.github/workflows/nix-build.yaml
```diff
@@ -24,7 +29,11 @@ jobs:
           extraPullNames: hermes-sdk,cosmos-nix
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - name: Run Nix build
+      - name: Remove Homebrew
+        if: ${{ matrix.os == 'macos-14' }}
         run: |
           sudo rm -rf /opt/homebrew
```
rnbguy, line 35: @Farhad-Shabani since you use macos, do we need this step?

PR #133, opened 2024-12-11
feat: integrate and test `{Conn,Chan}Open{Try,Confirm}` messages and events

### LINKED ISSUE #125: Modify integration tests to work with Cairo IBC core contract
Instead of interating with the Comet and ICS20 contracts directly, we would deploy the IBC core contract and interact via that.

rnbguy:
Tests the full connection and channel handshake at starknet.

Currently, it submits all the handshake messages at starknet. So it is starknet <> starknet handshake relay. This way, we test all the cairo handlers and events in e2e setup.

closes #125

### DIFF
--- a/cairo-contracts/packages/core/src/connection/components/handler.cairo
+++ b/cairo-contracts/packages/core/src/connection/components/handler.cairo
@@ -337,6 +337,7 @@ pub mod ConnectionHandlerComponent {
                     conn_end_on_b.client_id.clone(),
                     msg.conn_id_on_b.clone(),
                     conn_end_on_b.counterparty.client_id.clone(),
+                    // TODO(rano): this should conn_id_on_a
                     msg.conn_id_on_b
                 );
         }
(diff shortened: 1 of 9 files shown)


cairo-contracts/packages/core/src/connection/components/handler.cairo
```diff
@@ -337,6 +337,7 @@ pub mod ConnectionHandlerComponent {
                     conn_end_on_b.client_id.clone(),
                     msg.conn_id_on_b.clone(),
                     conn_end_on_b.counterparty.client_id.clone(),
+                    // TODO(rano): this should conn_id_on_a
```
rnbguy, line 340: @soareschen do you also think this is a bug?

cairo-contracts/packages/core/src/connection/components/handler.cairo
```diff
@@ -337,6 +337,7 @@ pub mod ConnectionHandlerComponent {
...
                     msg.conn_id_on_b
```
rnbguy, line 341: @soareschen do you also think this is a bug?

rnbguy: 8183e02

soareschen (approved):

PR #149, opened 2024-12-18
Scaffold channel handshake components

soareschen:


relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/channel_message.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct BuildStarknetToCosmosChannelHandshakeMessage;
+
+impl<Chain, Counterparty> ChannelOpenTryMessageBuilder<Chain, Counterparty>
+    for BuildStarknetToCosmosChannelHandshakeMessage
+where
+    Chain: HasMessageType + HasPortIdType<Counterparty> + HasErrorType,
+    Counterparty:
+        HasChannelIdType<Chain> + HasPortIdType<Chain> + HasChannelOpenTryPayloadType<Chain>,
+{
+    async fn build_channel_open_try_message(
```
rnbguy, line 20: Why there is no `ChannelOpenInitMessageBuilder` for `StarknetToCosmos` ?

soareschen: It is because we can use the existing version in cosmos-chain-components, since it does not depend on the counterparty types.

soareschen: https://github.com/informalsystems/ibc-starknet/blob/16473557efb1d0581c1bd13570a47cf434a98892/relayer/crates/starknet-chain-components/src/components/starknet_to_cosmos.rs#L51-L53

Issue #141, opened 2024-12-17
Introduce `StarknetMessage` type containing counterparty Cosmos height


soareschen:
Currently, we use `Call` as the Starknet message type. However when relaying IBC messages, the message needs to contain the counterparty chain height.

We will introduce a `StarknetMessage` type that wraps around `Call`, and also contains an `Option<CosmosHeight>`. We will then update the `CounterpartyMessageHeightGetter` implementation to return the counterparty Cosmos height.

PR #162, opened 2024-12-24
feat(e2e): bootstrap connection and channel between starknet <> cosmos

rnbguy:


rnbguy:
I get the following error when trying to bootstrap connection


Error: Status { code: Unknown, message: "unable to resolve type URL : tx parse error [cosmos/cosmos-sdk@v0.50.7/x/auth/tx/decoder.go:44] with gas used: '1174'", metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "67"} }, source: None }

rnbguy:
Currently, connection relay fails as the cosmos payload uses the latest height on Cosmos. Unless we update the light client on Starknet to that height, the conn_open_ack message on Starknet will fail, as it is not aware of the consensus_root for that corresponding height.

#141 is required to fix this.

rnbguy:
Currently, the test can't find Channel handshake events, as #135 didn't return `ChannelId` from channel message handlers.

https://github.com/informalsystems/ibc-starknet/blob/a151c39f8c899d7dd6c8438ff89509f78e6d33ec/cairo-contracts/packages/core/src/channel/components/handler.cairo#L210-L212

https://github.com/informalsystems/ibc-starknet/blob/a151c39f8c899d7dd6c8438ff89509f78e6d33ec/cairo-contracts/packages/core/src/channel/components/handler.cairo#L301-L303

Issue #171, opened 2025-01-07
Add integration test for ICS-20 from Starknet to Cosmos


Farhad-Shabani:
Add an integration test to validate the `send_transfer` functionality of ICS-20 Cairo contract. This involves implementing a test for token transfers from Starknet to Cosmos, complementing the existing `test_starknet_ics20_contract`, which already covers transfers in the Cosmos to Starknet direction.

The test should cover transferring a native token (e.g. STRK) to the Cosmos chain (escrow) and sending an IBC-created token back to the Cosmos chain (burn). It should also include acknowledging the packet on the Starknet node. To achieve this, the test can be implemented as either separate tests or a single comprehensive test.

Issue #150, opened 2024-12-19
When possible, use IBC types from ibc-rs


soareschen:
We should reuse types like `ClientId`, `ConnectionId`, `Packet`, etc. That would make it significantly easier to reuse the Cosmos implementations, which assume that the counterparty has these types. Otherwise, we have to copy over the Cosmos implementations, and perform minor tweaks to convert between the types.

The way we implement Cairo encoding means that we can easily implement them for non-owned types from ibc-rs. The main difference is that we will need to implement `FieldGetter` for the ibc-rs types, and then use encoders like `EncodeFieldWithGetter` to pass in the field getters explicitly.

rnbguy:
I will list down some examples to get this resolves quickly.

PR #172, opened 2025-01-08
feat(e2e): native token transfer from cosmos to starknet

### LINKED ISSUE #170: Enhance `test_starknet_ics20_contract` to fully simulate relaying from Cosmos to Starknet
## Context
In the [`test_starknet_ics20_contract`](https://github.com/informalsystems/ibc-starknet/blob/d542304d27902c00ba03d9df87864810e6eb4833/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L57) test, once the Cosmos and Starknet chains are connected (via the handshake process), a [dummy](https://github.com/informalsystems/ibc-starknet/blob/d542304d27902c00ba03d9df87864810e6eb4833/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L352) `msg_recv_packet` is created. To fully simulate the relaying of a packet from Cosmos to Starknet, the `msg_recv_packet` should be constructed based on a previously submitted `MsgTransfer` to the Cosmos chain.

Additionally, the packet should be acknowledged on the Cosmos chain and should cover two scenarios: first, when the incoming token into Starknet is an IBC-minted token, and second, when it is a native token previously sent to Cosmos and now being unescrowed.

rnbguy:
Partially resolves #170 #171

ics20 transfer: cosmos -> starknet -> cosmos

- [x] mint on starknet
- [x] burn on starknet

### DIFF
--- a/cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
+++ b/cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
@@ -25,5 +25,8 @@ pub trait ITransferQuery<TContractState> {
     /// ```
     /// Hashing the denom is delegated to the client as it is more cost-efficient.
     fn ibc_token_address(self: @TContractState, token_key: felt252) -> ContractAddress;
+
+    /// Return the contract addresses of all IBC tokens.
+    fn ibc_token_addresses(self: @TContractState) -> Array<ContractAddress>;
 }
 
--- a/cairo-contracts/packages/core/src/channel/types.cairo
+++ b/cairo-contracts/packages/core/src/channel/types.cairo
@@ -30,8 +30,9 @@ pub impl PacketImpl of PacketTrait {
 
     /// Checks if the packet is timed out.
     fn is_timed_out(self: @Packet, latest_height: @Height, latest_timestamp: @Timestamp) -> bool {
-        !(self.timeout_height_on_b > latest_height
-            && self.timeout_timestamp_on_b > latest_timestamp)
+        !((self.timeout_height_on_b.is_zero() || self.timeout_height_on_b > latest_height)
+            && (self.timeout_timestamp_on_b.is_zero()
+                || self.timeout_timestamp_on_b > latest_timestamp))
     }
 }
 
--- a/relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
@@ -22,7 +22,9 @@ use hermes_chain_type_components::traits::types::address::HasAddressType;
 use hermes_encoding_components::traits::encode::CanEncode;
 use hermes_encoding_components::traits::has_encoding::HasEncoding;
 use hermes_encoding_components::traits::types::encoded::HasEncodedType;
+use ibc::apps::transfer::types::packet::PacketData as IbcIcs20PacketData;
 use ibc::core::channel::types::packet::Packet as IbcPacket;
+use ibc::core::channel::types::timeout::{TimeoutHeight, TimeoutTimestamp};
 use ibc::core::client::types::Height;
 use starknet::accounts::Call;
 use starknet::core::types::Felt;
@@ -31,6 +33,10 @@ use starknet::macros::selector;
 use crate::impls::types::message::StarknetMessage;
 use crate::traits::queries::address::CanQueryContractAddress;
 use crate::types::cosmos::height::Height as CairoHeight;
+use crate::types::messages::ibc::denom::{Denom, PrefixedDenom};
+use crate::types::messages::ibc::ibc_transfer::{
+    IbcTransferMessage as CairoIbcTransferMessage, Participant,
+};
 use crate::types::messages::ibc::packet::{
     Acknowledgement as CairoAck, MsgAckPacket, MsgRecvPacket, MsgTimeoutPacket,
     Packet as CairoPacket, Sequence, StateProof,
@@ -53,7 +59,9 @@ where
             Chain,
             ReceivePacketPayload = ReceivePacketPayload<Counterparty>,
         >,
-    Encoding: CanEncode<ViaCairo, MsgRecvPacket> + HasEncodedType<Encoded = Vec<Felt>>,
+    Encoding: CanEncode<ViaCairo, MsgRecvPacket>
+        + CanEncode<ViaCairo, CairoIbcTransferMessage>
+        + HasEncodedType<Encoded = Vec<Felt>>,
 {
     async fn build_receive_packet_message(
         chain: &Chain,
@@ -71,7 +79,7 @@ where
         };
 
         let receive_packet_msg = MsgRecvPacket {
-            packet: CairoPacket::from(packet.clone()),
+            packet: from_cosmos_to_cairo_packet(packet, chain.encoding()),
             proof_commitment_on_a,
             proof_height_on_a,
         };
@@ -110,7 +118,9 @@ where
         + HasHeightType<Height = Height>
         + HasCommitmentProofType
         + HasAcknowledgementType<Chain, Acknowledgement = Vec<u8>>,
-    Encoding: CanEncode<ViaCairo, MsgAckPacket> + HasEncodedType<Encoded = Vec<Felt>>,
+    Encoding: CanEncode<ViaCairo, MsgAckPacket>
+        + CanEncode<ViaCairo, CairoIbcTransferMessage>
+        + HasEncodedType<Encoded = Vec<Felt>>,
 {
     async fn build_ack_packet_message(
         chain: &Chain,
@@ -128,7 +138,7 @@ where
         };
 
         let ack_packet_msg = MsgAckPacket {
-            packet: CairoPacket::from(packet.clone()),
+            packet: from_cosmos_to_cairo_packet(packet, chain.encoding()),
             acknowledgement: CairoAck {
                 ack: counterparty_payload.ack,
             },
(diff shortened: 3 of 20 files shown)


rnbguy:
Current test fails because of cosmos/ibc-rs#1262.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -330,188 +341,54 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+            address_starknet_b,
+            &Amount::new(transfer_quantity, denom_cosmos.clone()),
+            &None,
+        )
+        .await?;
 
-            let ibc_transfer_events_2: Vec<IbcTransferEvent> =
-                event_encoding.filter_decode_events(&response.events)?;
+        // TODO(rano): how do I get the ics20 token contract address from starknet events
+        cosmos_to_starknet_relay.relay_packet(&packet).await?;
```
rnbguy, line 374: We can use this:
https://github.com/informalsystems/ibc-starknet/blob/717b22900de39b7efee65f10239da9aee56b248f/cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo#L13-L27

rnbguy: Ended up with implementing a query endpoint that returns all token addresses, which is just one in the e2e test.

rnbguy:
Currently, the test fails, as the Starknet doesn't accept packet_ack because of commitment mismatch.

cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
```diff
@@ -25,5 +25,8 @@ pub trait ITransferQuery<TContractState> {
     /// ```
     /// Hashing the denom is delegated to the client as it is more cost-efficient.
     fn ibc_token_address(self: @TContractState, token_key: felt252) -> ContractAddress;
+
+    /// Return the contract addresses of all IBC tokens.
+    fn ibc_token_addresses(self: @TContractState) -> Array<ContractAddress>;
 }
```
Farhad-Shabani, line 31: If the goal of adding this endpoint is only to obtain the token address of an IBC-created token, it’s actually unnecessary. We can either listen to the `CreateTokenEvent` that gets emitted or construct the token's key on the relayer side and use the `ibc_token_address`, same as it's [computed](https://github.com/informalsystems/ibc-starknet/blob/9694808e28d4cb963abe6a68ee30bfa4d2c1d645/cairo-contracts/packages/apps/src/transfer/types.cairo#L120-L130) by the contract.

cairo-contracts/packages/core/src/channel/types.cairo
```diff
@@ -30,8 +30,9 @@ pub impl PacketImpl of PacketTrait {
 
     /// Checks if the packet is timed out.
     fn is_timed_out(self: @Packet, latest_height: @Height, latest_timestamp: @Timestamp) -> bool {
-        !(self.timeout_height_on_b > latest_height
-            && self.timeout_timestamp_on_b > latest_timestamp)
+        !((self.timeout_height_on_b.is_zero() || self.timeout_height_on_b > latest_height)
+            && (self.timeout_timestamp_on_b.is_zero()
+                || self.timeout_timestamp_on_b > latest_timestamp))
```
Farhad-Shabani, line 35: Nice catch!

relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
```diff
@@ -221,3 +229,116 @@ where
         Ok(message)
     }
 }
+
+fn from_cosmos_to_cairo_packet<Encoding>(packet: &IbcPacket, encoding: &Encoding) -> CairoPacket
+where
```
Farhad-Shabani, line 234: This conversion feels quite hacky, doesn’t seem worth investing in, especially since we’ll be implementing #150 soon. I’m becoming more convinced to prioritize #150 first (at least for the types used in this PR), then get back here. What do you think?

relayer/crates/starknet-chain-components/src/types/messages/ibc/ibc_transfer.rs
```diff
@@ -29,12 +35,46 @@ pub type EncodeIbcTransferMessage = CombineEncoders<
     ],
 >;
 
+#[derive(HasField)]
+pub struct IbcTransferSendMessage {
+    pub port_id_on_a: PortId,
+    pub chan_id_on_a: ChannelId,
+    pub packet_data: IbcTransferMessage,
```
Farhad-Shabani, line 42: What about renaming its type to something makes more sense? e.g. `IbcTransferPacketData`?

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -330,188 +360,239 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+            &cosmos_channel_id,
+            &IbcPortId::transfer(),
+            wallet_cosmos_a,
+            address_starknet_b,
+            &Amount::new(transfer_quantity, denom_cosmos.clone()),
+            &None,
+        )
+        .await?;
+
+        // TODO(rano): how do I get the ics20 token contract address from starknet events
```
Farhad-Shabani, line 392: When a token is created, the ICS20 emits a `CreateTokenEvent` that includes the token's address. We can read the address by retrieving that event.

Farhad-Shabani (changes requested):

cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
rnbguy: It is indeed unnecessary. I added it as a hack. I just forgot to add a `TODO` comment on top of it.

Essentially, the `relay_packet` method should return any stateful data generated during the relay -- specifically token address of the minted denom. I will wait for Soares to fix this. After that, we can remove this.

relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
rnbguy: I would like to finish up the handshake and relays, as I already started working on this.

Or, we can merge until here and work on #150 and then resume work on getting packet relay working.

relayer/crates/starknet-chain-components/src/types/messages/ibc/ibc_transfer.rs
rnbguy: I used `IbcTransferSendMessage` as it is a message that is submitted to the ics20 contract. But I can refactor the names more sensibly and consistent with Cairo counterparts.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
rnbguy: The comment is mainly about how to use `relay_packet` and still get the token address. The hack, that you pointed in https://github.com/informalsystems/ibc-starknet/pull/172#discussion_r1907728623, is introduced exactly for this.

rnbguy:
I wanted to perform erc20 token transfer from starknet to cosmos, but I realized, we need to `permit` the ics20 contract on behalf of the sender. I am not sure how to do that.

I am making this PR final. Let's test starknet -> cosmos -> starknet on a different PR.

relayer/crates/starknet-chain-components/src/types/messages/ibc/ibc_transfer.rs
rnbguy: 9881575

cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
Farhad-Shabani: Could you open up an issue for this? What exactly is blocking us and what's needed in `hermes-sdk` so that we can rely on existing facilities here?

relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
Farhad-Shabani: I think we have enough time ahead of us, thus not recommend leaving the PR with this function. 
Let's get #150 done first. Let me know if I can be assist on that.

Farhad-Shabani:
@rnbguy Now that the test is working, we can be confident in the correctness of our implementation on both the relayer and contracts side. My suggestion is to address https://github.com/informalsystems/ibc-starknet/pull/172#discussion_r1909003884 and https://github.com/informalsystems/ibc-starknet/pull/172#discussion_r1907729760 first, update the PR, and then wrap it up.

Farhad-Shabani:
> I wanted to perform erc20 token transfer from starknet to cosmos, but I realized, we need to permit the ics20 contract on behalf of the sender. I am not sure how to do that.

The way it works at the moment, users should approve their allowance before initiating a `send_transfer` like it's done [here](https://github.com/informalsystems/ibc-starknet/blob/04c89882934470e9d9110d093c25c31495f28fea/cairo-contracts/packages/contracts/src/tests/transfer.cairo#L21-L22) in the contract's unit tests.

Farhad-Shabani (approved):
By the way, I approve the PR, leaving that to you how to organize issues/PRs moving forward.

cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo
rnbguy: #175

relayer/crates/starknet-chain-components/src/impls/messages/packet.rs
rnbguy: Since, I am wrapping up the relay, is it okay if I leave it as it is. The above concern is already mentioned in #150, so I am not open a new issue.

Issue #147, opened 2024-12-17
Store chain ID in both Starknet and Comet client state


soareschen:
The client states should contain the chain ID.

After updating the fields, the `ClientStateFieldsGetter` implementation should be updated accordingly.

PR #173, opened 2025-01-08
Add ChainId to Starknet and Comet client state

ljoss17:
Closes: #147

### Status

* [x] Added `chain_id` field to `CometClientState` in relayer code
* [x] Added `chain_id` field to `CometClientState` in Cairo code
* [x] Added `chain_id` field to `StarknetClientState` in relayer code


light-client/ibc-client-starknet-types/src/encoding/impls/client_state.rs
```diff
@@ -21,22 +22,27 @@ delegate_components! {
...
 impl Transformer for EncodeStarknetClientState {
-    type From = Product![Height];
+    type From = Product![Height, Felt];
 
     type To = StarknetClientState;
 
-    fn transform(product![latest_height]: Self::From) -> Self::To {
-        StarknetClientState { latest_height }
+    fn transform(product![latest_height, chain_id]: Self::From) -> Self::To {
+        StarknetClientState { latest_height, chain_id: chain_id.to_string() }
```
rnbguy, line 46: This should be the same way we print `ClientId` with `client_type` field with `Felt` type.

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/starknet-chain-components/src/types/client_id.rs#L43-L50

relayer/crates/starknet-chain-components/src/types/cosmos/client_state.rs
```diff
@@ -173,3 +181,33 @@ impl From<CometClientState> for Any {
...
+    Encoding: CanDecodeMut<Strategy, String> + CanRaiseError<&'static str>,
+{
+    fn decode_mut<'a>(
+        encoding: &Encoding,
+        buffer: &mut Encoding::DecodeBuffer<'a>,
+    ) -> Result<ChainId, Encoding::Error> {
+        let chain_id_str = encoding.decode_mut(buffer)?;
+        ChainId::new(&chain_id_str).map_err(|_| Encoding::raise_error("invalid chain id"))
+    }
+}
```
rnbguy, line 213: Is this correct to satisfy `CanEncodeAndDecodeMut<ViaProtobuf, ChainId>`?

light-client/ibc-client-starknet-types/src/encoding/impls/client_state.rs
```diff
@@ -21,22 +22,30 @@ delegate_components! {
                     symbol!("latest_height"),
                     EncodeLengthDelimitedProtoField<1, UseContext>,
                 >,
+                // EncodeField<
+                //     symbol!("chain_id"),
+                //     EncodeLengthDelimitedProtoField<2, UseContext>,
+                // >,
```
rnbguy, line 28: The commented out part are done the same way as [`EncodeStarknetConsensusState`](https://github.com/informalsystems/ibc-starknet/blob/848eab5fa9d86caf6c1bf48b6ea983f72c90f52f/light-client/ibc-client-starknet-types/src/encoding/impls/consensus_state.rs) and [`EncodeStarknetHeader`](https://github.com/informalsystems/ibc-starknet/blob/848eab5fa9d86caf6c1bf48b6ea983f72c90f52f/light-client/ibc-client-starknet-types/src/encoding/impls/header.rs). Not sure why this doesn't work.

relayer/crates/starknet-chain-components/src/types/cosmos/client_state.rs
soareschen: I have explained this to @ljoss17. This encoding strategy works for Cairo, but not for protobuf. This is because protobuf requires special field tag to be applied to field values, and "primitive" values like strings can only be field values, meaning they always require a tag.

The solution is to define a `EncodeChainIdField<const TAG: u32>` encoder, which makes use of `EncodeStringField<TAG>` to encode the given protobuf field.

soareschen:
@ljoss17 will move the `EncodeChainIdField` encoder to Hermes SDK, and then we will use the construct here once the corresponding PR is merged.

ljoss17:
> @ljoss17 will move the `EncodeChainIdField` encoder to Hermes SDK, and then we will use the construct here once the corresponding PR is merged.

Opened https://github.com/informalsystems/hermes-sdk/pull/525

light-client/ibc-client-starknet/src/client_state/execution.rs
```diff
@@ -61,6 +61,7 @@ where
 
         let new_client_state = ClientStateType {
             latest_height: header.height,
+            chain_id: "dummy".parse()?, // TODO: fill with correct info
```
soareschen, line 64: I think we need to query the current client state here, and then use the chain ID from the original client state.

ljoss17: Done in 279deb4858789e3726fa5eaf7af36d2cb65af870

soareschen (approved):

Issue #174, opened 2025-01-09
Develop reusable setups for relayer's integration tests


Farhad-Shabani:
## Context
Now that the full set of IBC layers has been implemented, the relayer's integration tests, specifically those under `ics20.rs` (e.g. [here](https://github.com/informalsystems/ibc-starknet/blob/d542304d27902c00ba03d9df87864810e6eb4833/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L57)), involve multiple steps like contract declaration, deployments, handshakes, and client creations, and so on, which are mostly repetitive across tests. Thus it makes sense to develop reusable testing artifacts. We can streamline the setup process, allowing necessary artifacts to be initialized in just a few lines. This approach will make it easier to focus on writing tests.

A similar method has already been applied to [Cairo's contract tests](https://github.com/informalsystems/ibc-starknet/blob/04c89882934470e9d9110d093c25c31495f28fea/cairo-contracts/packages/contracts/src/tests/transfer.cairo#L11-L15), which could serve as a reference.

Caught my eye while reviewing #172.

rnbguy:
Totally agree. IMO, it's best to implement these as hermes-sdk components.

Farhad-Shabani:
@soareschen, could you hash this out as an issue in the Hermes SDK, if that makes sense?

soareschen:
The whole thing can be simplified by building a binary channel test driver, similar to what we did in Hermes SDK. After that, we can work on making it work with the abstract IBC test suite `TestIbcTransfer`.

PR #176, opened 2025-01-10
feat(e2e): transfer erc20 token from starknet to cosmos and back

### LINKED ISSUE #170: Enhance `test_starknet_ics20_contract` to fully simulate relaying from Cosmos to Starknet

### LINKED ISSUE #171: Add integration test for ICS-20 from Starknet to Cosmos

rnbguy:
Closes #170
Closes #171

ics20 transfer: starknet -> cosmos -> starknet

- [x] escrow on starknet
- [x] unescrow on starknet

### DIFF
--- a/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
+++ b/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
@@ -872,6 +872,31 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
 
         cosmos_to_starknet_relay.relay_packet(&packet).await?;
 
+        let balance_cosmos_a_step_4 = cosmos_chain
+            .query_balance(address_cosmos_a, &cosmos_ibc_denom)
+            .await?;
+
+        info!(
+            "cosmos balance after transfer back to starknet: {}",
+            balance_cosmos_a_step_4
+        );
+
+        assert_eq!(balance_cosmos_a_step_4.quantity, 0u64.into());
+
+        let balance_starknet_relayer_step_4 = starknet_chain
+            .query_token_balance(&erc20_token_address, address_starknet_relayer)
+            .await?;
+
+        info!(
+            "starknet balance after transfer back to starknet: {}",
+            balance_starknet_relayer_step_4
+        );
+
+        assert_eq!(
+            balance_starknet_relayer_step_4.quantity,
+            balance_starknet_relayer_step_0.quantity
+        );
+
         Ok(())
     })
 }
--- a/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
+++ b/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
@@ -776,7 +776,7 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
 
         // create ibc packet
 
-        let _starknet_ibc_packet = {
+        let starknet_ibc_packet = {
             let timeout_height_on_b = IbcHeight::new(
                 send_packet_event.timeout_height_on_b.revision_number,
                 send_packet_event.timeout_height_on_b.revision_height,
@@ -813,6 +813,12 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
                 timeout_timestamp_on_b,
             }
         };
+
+        // relay the packet via starknet to cosmos relay
+
+        starknet_to_cosmos_relay
+            .relay_packet(&starknet_ibc_packet)
+            .await?;
 
         Ok(())
     })
--- a/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
+++ b/relayer/crates/starknet-integration-tests/src/tests/ics20.rs
@@ -11,7 +11,9 @@ use hermes_cosmos_chain_components::types::connection::CosmosInitConnectionOptio
 use hermes_cosmos_integration_tests::init::init_test_runtime;
 use hermes_cosmos_relayer::contexts::build::CosmosBuilder;
 use hermes_cosmos_relayer::contexts::chain::CosmosChain;
+use hermes_cosmos_test_components::chain::impls::transfer::amount::derive_ibc_denom;
 use hermes_cosmos_test_components::chain::types::amount::Amount;
+use hermes_cosmos_test_components::chain::types::denom::Denom as IbcDenom;
 use hermes_cosmos_wasm_relayer::context::cosmos_bootstrap::CosmosWithWasmClientBootstrap;
 use hermes_encoding_components::traits::decode::CanDecode;
 use hermes_encoding_components::traits::encode::CanEncode;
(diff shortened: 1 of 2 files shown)


relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -589,16 +595,308 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                fixed_supply: erc20_token_supply.into(),
+                recipient: *address_starknet_relayer,
+                owner: *address_starknet_relayer,
+            };
+
+            let calldata = StarknetCairoEncoding.encode(&deploy_message)?;
+
+            starknet_chain
+                .deploy_contract(&erc20_class_hash, false, &calldata)
+                .await?
```
Farhad-Shabani, line 628: Can’t we use the [pre-deployed token contracts](https://0xspaceshard.github.io/starknet-devnet-rs/docs/balance) on devnet? That way, we wouldn’t need to deploy another token. This one though can be a good another testing scenario, but maybe not the primary use case.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -589,16 +595,308 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+        //     info!("response: {:?}", response);
+
+        //     let balance = starknet_chain
+        //         .query_token_balance(&erc20_token_address, address_starknet_b)
+        //         .await?;
+
+        //     info!("top-up balance: {}", balance);
+
+        //     assert_eq!(balance.quantity, transfer_quantity.into());
+        // }
```
Farhad-Shabani, line 668: Should this be removed?

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -589,16 +595,308 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                memo,
+            }
+        };
+
+        // create ibc transfer message
+
+        let starknet_ics20_send_message = {
+            let current_starknet_time = SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)?
+                .as_secs();
```
Farhad-Shabani, line 720: This isn’t the actual current Starknet time. Right? might be confusing. Couldn’t query the node for its timestamp?

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -589,16 +595,308 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                seq_on_a: send_packet_event.sequence_on_a.sequence.into(),
+                port_id_on_a: send_packet_event.port_id_on_a.port_id.parse()?,
+                chan_id_on_a: send_packet_event.channel_id_on_a.channel_id.parse()?,
+                port_id_on_b: send_packet_event.port_id_on_b.port_id.parse()?,
+                chan_id_on_b: send_packet_event.channel_id_on_b.channel_id.parse()?,
+                data: serde_json::to_vec(&ibc_transfer_packet_data).unwrap(),
+                timeout_height_on_b,
+                timeout_timestamp_on_b,
+            }
+        };
```
Farhad-Shabani, line 817: I wonder if `hermes-sdk` does not have a component that listens to send transfer events, handles these underlying insertions, and packet creation? Should we open an issue to add this to the hermes-sdk (or maybe the relayer, not entirely sure)?

Farhad-Shabani (commented):
Thanks Rano for the tests!
Left a few comments and a general suggestion: it’d be better if we split each direction into its own test case. It’ll make debugging way easier later on. 
By the way, I opened issue #174, perhaps using a reusable test setup should make it simple to separate the tests.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
rnbguy: bf09dbe

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
rnbguy: bf09dbe

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
rnbguy: c8b370c

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
rnbguy: Yes, hermes-sdk can support this. I am using the `CanIbcTransferToken<CounterpartyChain>::ibc_transfer_token` when transfering token from cosmos-sdk to starknet.

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L381-L393

But I don't know how to implement it for starknet to cosmos. We need to the following trait bound to be satisfied,

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/starknet-chain-context/src/contexts/chain.rs#L367-L368

rnbguy: opened #181

Farhad-Shabani: I get a bit worried about TODOs in the code. They’re like socks in the laundry, just disappearing forever!
Let’s create issues instead of leaving them there. To keep things simple, the issues don’t have to be super detailed. Just a short, straightforward one is fine unless extra details are really necessary.

Farhad-Shabani: Got it, thanks for pointing that out!

Farhad-Shabani: Oh, I see you already opened up a bunch. Much appreciated!

Farhad-Shabani (approved):
🎉

PR #186, opened 2025-01-11
feat(relayer): implement poseidon hash

### LINKED ISSUE #175: query minted token address on starknet using static keys
hermes-sdk has [derive_ibc_denom](https://github.com/informalsystems/hermes-sdk/blob/3119fa1bc7b3b8af53f9680f54909ae0a4a1efbe/crates/cosmos/cosmos-test-components/src/chain/impls/transfer/amount.rs#L46-L50) to find the ibc minted denoms (of the form `ibc/<hash>`) in cosmos-sdk based chains. This is a pure function as `<hash>` can be derived statically from trace prefix and native denom.

But in starknet, the tokens are minted by deploying a new erc20 contract. So the minted token denoms (contract addresses) can't be derived statically. The ics20 contract maintains a map of keys (which can be statically derived) of the minted token and the corresponding token addresses.

https://github.com/informalsystems/ibc-starknet/blob/717b22900de39b7efee65f10239da9aee56b248f/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L34-L39

Thus, `StarknetChain` needs a method which takes the same arguments as `derive_ibc_denom` and derive the key (using poseidon hash) and query the ics20 contract for the token address.

https://github.com/informalsystems/ibc-starknet/blob/717b22900de39b7efee65f10239da9aee56b248f/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L777-L781

https://github.com/informalsystems/ibc-starknet/blob/717b22900de39b7efee65f10239da9aee56b248f/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L276-L284

### LINKED ISSUE #180: [relayer] poseidon hash of serialization of cairo types
We need to implement poseidon hash on `Vec<Felt>`.

### Context
ics20 contract uses poseidon hash to hash the ibc prefixed denoms and use it as keys to maintain a map between ibc prefixed denoms and erc20 contract addresses. We need this hash to query erc20 contract addresses after relaying a denom from cosmos.

more at #175

rnbguy:
Closes #175
Closes #180


cairo-contracts/packages/apps/src/transfer/types.cairo
```diff
@@ -120,11 +120,12 @@ impl PrefixedDenomDisplay of Display<PrefixedDenom> {
...
-        while let Option::Some(path) = trace_path_span.pop_front() {
-            key_builder.append_serde(path);
-        };
-        key_builder.append_serde(self.base);
+        // let mut trace_path_span = self.trace_path.span();
+        // while let Option::Some(path) = trace_path_span.pop_front() {
+        //     key_builder.append_serde(path);
+        // };
+        // key_builder.append_serde(self.base);
+        key_builder.append_serde(self);
```
rnbguy, line 128: I am not sure why `PrefixedDenom` is hashed manually when it implements `Serde` already.

I am ok to revert this and make it compatible at relayer e2e side if this is necessary.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -409,22 +410,33 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                }],
+                base: Denom::Hosted(denom_cosmos.to_string()),
+            };
+
+            // https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/cairo-contracts/packages/utils/src/utils.cairo#L36
+            let ibc_prefixed_denom_key = PoseidonState::default()
+                .update(PoseidonState::update_slice(
+                    &cairo_encoding.encode(&ibc_prefixed_denom)?,
+                ))
+                .finish();
```
rnbguy, line 427: The `.key()` method in Cairo uses,

```rs
PoseidonTrait::new().update(poseidon_hash_span(self.data.span())).finalize()
```

That's why I had to use it this way. I am not sure, why the double hashing is necessary. Why can't we use the following?

```rs
poseidon_hash_span(self.data.span())
```

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -409,22 +410,31 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                    port_id: ics20_port.to_string(),
+                    channel_id: starknet_channel_id.channel_id.clone(),
+                }],
+                base: Denom::Hosted(denom_cosmos.to_string()),
+            };
+
+            // https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/cairo-contracts/packages/utils/src/utils.cairo#L36
+            let ibc_prefixed_denom_key = PoseidonState::digest(&[PoseidonState::digest(
+                &cairo_encoding.encode(&ibc_prefixed_denom)?,
+            )]);
```
rnbguy, line 425: The `.key()` method in Cairo uses,

https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/cairo-contracts/packages/utils/src/utils.cairo#L36

I am not sure why the double hashing is necessary. Why can't we use the following?

```rs
poseidon_hash_span(self.data.span())
```

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
```diff
@@ -409,22 +410,31 @@ fn test_starknet_ics20_contract() -> Result<(), Error> {
...
+                    port_id: ics20_port.to_string(),
+                    channel_id: starknet_channel_id.channel_id.clone(),
+                }],
+                base: Denom::Hosted(denom_cosmos.to_string()),
+            };
+
+            // https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/cairo-contracts/packages/utils/src/utils.cairo#L36
+            let ibc_prefixed_denom_key = Poseidon3Hasher::digest(&[Poseidon3Hasher::digest(
+                &cairo_encoding.encode(&ibc_prefixed_denom)?,
+            )]);
```
rnbguy, line 425: The `.key()` method in Cairo uses,

https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/cairo-contracts/packages/utils/src/utils.cairo#L36

I am not sure why the double hashing is necessary. Why can't we use the following?

```rs
poseidon_hash_span(self.data.span())
```

cairo-contracts/packages/apps/src/transfer/types.cairo
rnbguy: I see. But why is this necessary - having the same key for base denom and prefixed denom with empty prefix ?

We are indexing minted tokens. Their prefixes are always non-empty.

relayer/crates/starknet-integration-tests/src/tests/ics20.rs
Farhad-Shabani: Yeah, this is wrong. I think the`poseidon_hash_span` naming threw me off, Anyway, I opened a PR: https://github.com/informalsystems/ibc-starknet/pull/189

Farhad-Shabani (commented):
It's interesting that there's no reference Rust implementation to use instead, so we wouldn't have to deal with these low-level details.
Anyway, great work here. Looks awesome to me! 
Once everything is finalized, I'll give it one last review, and approve. Thx!

cairo-contracts/packages/apps/src/transfer/types.cairo
```diff
@@ -120,6 +120,17 @@ impl PrefixedDenomDisplay of Display<PrefixedDenom> {
...
+        //
+        // We wanted to get the same hash for a standalone `base`
+        // and a `PrefixedDenom` the same `base` with empty `TracePath`.
+        //
+        // So, we pass serialization of each `TracePrefix` in `TracePath`
+        // separately and then the serialization of `base`.
+        //
+        // `key_builder.append_serde(self)` would have included the serialization
+        // of empty array which contains the length (zero).
+        // So, the hash would  have been different.
```
Farhad-Shabani, line 133: Thank you so much!

cairo-contracts/packages/utils/src/utils.cairo
```diff
@@ -59,3 +59,62 @@ pub impl RemotePathBuilderImpl of RemotePathBuilderTrait {
...
+        // depending on how you serialize the struct, the key will be different
+        let value = Foo { foo: array!["hello", "world"] };
+        assert_eq!(
+            direct_key(@value), 0x8a7f3b665d11363ad83b7fb1ee769c513497f2353a70250b0bddcaaaf6458b
+        );
+        assert_eq!(
+            manual_key(@value), 0x1241da65bd34c17206635225f7a86ab41089a0e14ced6dc24b3b831c78ff4b4
+        );
+    }
+}
```
Farhad-Shabani, line 120: Should we keep these tests? If I’m not mistaken, they were meant to demonstrate that manual key hashing doesn’t yield the same results as direct hashing. Though, it seems they don’t cover any actual implementation.

Farhad-Shabani (approved):
Left a comment which I'll let you decide. Otherwise all set. Thanks

cairo-contracts/packages/utils/src/utils.cairo
rnbguy: 4e47548

PR #197, opened 2025-01-14
feat(relayer): reuse `ConnectionEnd` and `ChannelEnd` from `ibc-rs`

### LINKED ISSUE #150: When possible, use IBC types from ibc-rs

rnbguy:
Use `ConnectionEnd` and `ChannelEnd` from `ibc-rs` and all other dependent types.

I am rexporting `ibc-rs` types as Starknet/Cairo types to avoid too many changes in the imports. So, you see `StarknetChannelId` and `IbcChannelId` (as similar) at some places. These are actually the same types. 

linked to #150

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/packet_fields.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/packet_fields.rs
@@ -14,8 +14,6 @@ where
         + HasChannelIdType<Counterparty, ChannelId = ChannelId>,
 {
     fn packet_src_channel_id(packet: &Packet) -> ChannelId {
-        ChannelId {
-            channel_id: packet.chan_id_on_b.to_string(),
-        }
+        packet.chan_id_on_b.clone()
     }
 }
--- a/relayer/crates/starknet-chain-components/src/types/client_id.rs
+++ b/relayer/crates/starknet-chain-components/src/types/client_id.rs
@@ -17,6 +17,7 @@ where
         value: &ClientId,
         buffer: &mut Encoding::EncodeBuffer,
     ) -> Result<(), Encoding::Error> {
+        // FIXME: add `sequence_number` method at `ibc-rs`
         let (client_type, sequence) = value.as_str().rsplit_once('-').expect("valid client id");
         let seq_u64 = sequence.parse::<u64>().expect("valid sequence");
         let client_type_felt = string_to_felt(client_type).expect("valid client type");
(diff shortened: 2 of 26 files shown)


rnbguy:
I realized that, changing `Has*Type` makes things not compile because of `hermes-sdk` wiring.

relayer/crates/starknet-chain-components/src/impls/packet_fields.rs
```diff
@@ -14,8 +14,6 @@ where
         + HasChannelIdType<Counterparty, ChannelId = ChannelId>,
 {
     fn packet_src_channel_id(packet: &Packet) -> ChannelId {
-        ChannelId {
-            channel_id: packet.chan_id_on_b.to_string(),
```
rnbguy, line 18: shouldn't this be `packet.chan_id_on_a`?

rnbguy:
Let's get this merged to avoid merge conflicts, as it became a significant change.

Note, I believe these are the only ones that are needed for membership verification (will be passed as protobuf serialized at wasm light client by `ibc-go`)

Apart from these types, we also need
- packet commitment
- packet ack
- packet receipt

But they are just `Vec<u8>`. So they are not really a risk for protobuf serialization.

relayer/crates/starknet-chain-components/src/types/client_id.rs
```diff
@@ -1,86 +1,41 @@
...
+where
+    Encoding: CanEncodeMut<Strategy, Product![Felt, u64]>,
+{
+    fn encode_mut(
+        encoding: &Encoding,
+        value: &ClientId,
+        buffer: &mut Encoding::EncodeBuffer,
+    ) -> Result<(), Encoding::Error> {
+        // FIXME: add `sequence_number` method at `ibc-rs`
+        let (client_type, sequence) = value.as_str().rsplit_once('-').expect("valid client id");
```
soareschen, line 21: Errors should be raised at least using `Encoding: CanRaiseError<String>`.

soareschen (approved):

Issue #164, opened 2025-01-04
Implement Merkle proof verification (existence & non-existence) for Comet contract


Farhad-Shabani:
## Summary

Implement the verification of existence (`verify_membership`) and non-existence (`verify_non_membership`) Merkle proofs for the CometBFT Cairo contract. These proofs typically accompany various IBC packets sent to a counterparty chain and are validated by the light client representing the source chain.

Farhad-Shabani:
## Protobuf Decoding: On-Chain vs. Off-Chain Approaches?

When we retrieve commitment proofs from Cosmos chains, they’re protobuf-encoded as a `u8` array. For Cairo contracts to verify these proofs, they need to be deserialized into Cairo-specific types. To make this happen, the data has to go through several conversions. Depending on whether Protobuf decoding happens on-chain or off-chain, there are two ways to handle it. Here’s how it breaks down, with roles split between the contract and the relayer:

### 1. Decoding On-Chain

1. **Relayer**: Fetch the proof as Protobuf-encoded `u8` bytes from the Cosmos chain.
2. **Relayer**: Pack the proof bytes into an IBC message and serialize it into a `felt252` array.
3. **Relayer**: Submit the message.
4. **Contract**: Deserialize the message and pull out the proof as `u8` bytes.
5. **Contract**: Protobuf decode the proof bytes into a Cairo type.

### 2. Decoding Off-Chain

1. **Relayer**: Fetch the proof as Protobuf-encoded `u8` bytes from the Cosmos chain.
2. **Relayer**: Protobuf decode the proof bytes into a domain Rust structure that mirrors its Cairo type.
3. **Relayer**: Pack the decoded proof into an IBC message and serialize it into a `felt252` array.
4. **Relayer**: Submit the message.
5. **Contract**: Deserialize the message and directly obtain the proof as a Cairo type.

## Proposal

**(1)** is the standard approach for IBC integrations, especially in Cosmos chains. But (**2)** is more cost-efficient since the heavy lifting (decoding) is done off-chain. That said, **(2)** does add a bit of extra work for the relayer. On the flip side, it keeps the contract logic simple, which makes it easier to maintain and boosts security.

PR #200, opened 2025-01-15
feat(contract): implement ICS23 vector commitment in Cairo (WIP)

Farhad-Shabani:
Part (1) of #164


cairo-libs/packages/ics23/src/types.cairo
```diff
@@ -0,0 +1,395 @@
...
+
+impl ProofAsProtoMessage of ProtoMessage<Proof> {
+    fn encode_raw(self: @Proof, ref context: EncodeContext) {
+        match self {
+            Proof::Exist(p) => p.encode_raw(ref context),
+            Proof::NonExist(p) => p.encode_raw(ref context),
+        }
+    }
+
+    fn decode_raw(ref self: Proof, ref context: DecodeContext) {
```
Farhad-Shabani, line 50: This is a temporary implementation until we appropriately handle #265.

cairo-libs/packages/cometbft/src/ics23.cairo
```diff
@@ -1,219 +0,0 @@
-use protobuf::types::message::{
```
Farhad-Shabani, line 1: Moved into the `packages/ics23` library, and the enums have been revised to include only the operations (variants) that are already supported on Starknet or those we have an implementation for.

cairo-libs/packages/protobuf/src/varint.cairo
```diff
@@ -0,0 +1,91 @@
+use protobuf::errors::ProtobufErrors;
```
Farhad-Shabani, line 1: I've gathered the `Varint` conversions that were added earlier, along with the new ones, under the `varint` module. Similarly, the related tests have been placed under `protobuf/src/tests/varint.cairo` for better organization.

cairo-libs/packages/protobuf/src/primitives/array.cairo
```diff
@@ -68,3 +66,26 @@ pub impl ArrayAsProtoMessage<T, +ProtoMessage<T>, +Drop<T>, +Default<T>> of Prot
     }
 }
 
+pub impl BytesAsProtoMessage of ProtoMessage<Array<u8>> {
```
Farhad-Shabani, line 69: More context on this: #267

cairo-libs/packages/ics23/src/types.cairo
```diff
@@ -0,0 +1,392 @@
...
+                        * 4 > *s.inner_spec.child_size && s.inner_spec.child_size >= @32),
+                    ICS23Errors::INVALID_INNER_SPEC,
+                );
+            }
+        };
+        hash
+    }
+}
+
+impl ExistenceProofAsProtoMessage of ProtoMessage<ExistenceProof> {
```
soareschen, line 103: Actually, I think we don't need to encode the Merkle proofs as protobuf. They are not part of the chain commitment, and thus can be freely re-encoded as Cairo when transmitted by the relayer.

cairo-libs/packages/ics23/Scarb.toml
```diff
@@ -0,0 +1,27 @@
...
+scarb-version = { workspace = true }
+authors       = { workspace = true }
+repository    = { workspace = true }
+license       = { workspace = true }
+Readme        = { workspace = true }
+keywords      = { workspace = true }
+
+description = """
+    Contains the implementation of ICS-23 vector commitments in Cairo.
+"""
```
rnbguy, line 15:
```suggestion
keywords      = { workspace = true }
description   = "Implementation of ICS-23 vector commitments in Cairo."
```

rnbguy (approved):
✨

cairo-libs/packages/ics23/src/types.cairo
```diff
@@ -0,0 +1,392 @@
...
+impl InnerOpAsProtoName of ProtoName<InnerOp> {
+    fn type_url() -> ByteArray {
+        "InnerOp"
+    }
+}
+
+#[derive(Default, Debug, Copy, Drop, PartialEq, Serde)]
+pub enum HashOp {
+    #[default]
+    NoOp,
```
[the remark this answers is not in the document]
rnbguy: Good point. but let's tackle the discussion in a different issue? Because it's present in `main` as part of the old protobuf impl done by me.

cairo-libs/packages/ics23/src/types.cairo
rnbguy: Good idea. But this requires further exploration into relayer code -- which is out of scope of this PR. Maybe we open an issue?

cairo-libs/packages/ics23/src/types.cairo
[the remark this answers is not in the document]
Farhad-Shabani: In production, only SHA256 is supported—anything else causes verification to fail. That said, I’d still advocate for keeping the `NoOp` variant. One really useful function of this variant that I found is in testing. It lets you skip hashing and check whether the verifier's behavior stays correct. (Defaults, No ops, zeros all should be rejected by verifier functions) 
Additionally, I think making `NoOp` as the default improves the system security, since if no proof spec or hash op is specified, it defaults to `NoOp`, causing proof verification to fail. Basically this forces users / relayers to be explicit about the hash ops in their proof specification.

cairo-libs/packages/ics23/src/types.cairo
Farhad-Shabani: I'm here basically implementing ICS-23 as a library. That's the next step. To decide how to integrate ICS-23 into both the Cairo and relayer codebases. We have the flexibility to use either encoded Protobuf data on-chain or the Cairo-encoded version.

By the way, I preferred to include the Protobuf implementation here as well, specifically since we were having bunch of artifacts ready to use. A main reason is that it allows me to reuse the existing `cosmos/ics23` [test data](https://github.com/cosmos/ics23/tree/master/testdata), keeping them as reference test cases. Otherwise, we’d have to generate new test data, which could raise questions about correctness.

PR #205, opened 2025-01-20
Verify Header in Wasm light client

ljoss17:
Closes: #216

Following #203 which added a `SignedStarknetHeader`, this PR adds a public key field to the client state and uses the public key to verifies the header and signature in the `SignedStarknetHeader`


light-client/ibc-client-starknet/src/client_state/execution.rs
```diff
@@ -56,7 +60,37 @@ where
...
+                description: "missing Deps from context".to_owned(),
+            })?;
+
+        match deps.api.secp256k1_verify(
+            header_digest.as_slice(),
+            &signed_header.signature,
+            &self.0.pub_key,
+        ) {
+            Ok(validation) if validation => {}
+            Ok(_) => {
```
rnbguy, line 79:
```suggestion
            Ok(true) => {}
            Ok(false) => {
```

ljoss17:
For `verify_membership_raw` and `verify_non_membership_raw`, https://github.com/informalsystems/ibc-starknet/blob/4c6324881ee89f8f191a108ff4c45a5c424b4b1e/light-client/ibc-client-starknet/src/client_state/common.rs#L50-L69, we don't have access to the `Context`, thus to the `secp256k1_verify` method. One solution would be to use the `secp256k1` library but this causes the Wasm contract to be too big to upload on Osmosis

soareschen (approved):

Issue #195, opened 2025-01-14
[cli] Implement create client command


soareschen:
We will use the `RunCreateClientCommand` implementation from Hermes SDK

PR #227, opened 2025-01-22
Implement `create client` CLI

seanchen1991:
Closes: #195


soareschen:
> @soareschen Does the `ClientSubCommand` need to be added as a variant of `AllSubCommands`?

Yeah we should add it there.

soareschen (approved):

PR #230, opened 2025-01-27
fix(lc/starknet): use `ibc-rs` patch for empty `revision_number` in `ibc-go-wasm-v8`

### LINKED ISSUE #233: [lc/starknet] `ibc-go-v8-wasm` fails with `ibc-rs`
`osmosis` testnet uses `ibc-go-v8` which doesn't use a fix that is included `ibc-go-v9` to work with `ibc-rs`

`osmosis` informed us they can't upgrade to `v9` because the upgrade will not be _state compatible_.

So, we need to patch the current `ibc-rs` to work with `v8`.

More details at cosmos/ibc-rs#1262

rnbguy:
closes #233

patch at ibc-rs: https://github.com/cosmos/ibc-rs/compare/rano/empty-rev-number-in-height

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -59,7 +59,7 @@
 
           rust-1_79 = nixpkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain-1.79.toml;
 
-          wasm-simapp = cosmos-nix.ibc-go-v9-wasm-simapp;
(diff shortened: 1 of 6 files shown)


rnbguy:
After this, we should switch to `osmosisd` binary for ics20 test.

Issue #206, opened 2025-01-20
Challenge in sending packets to ICS-20 contract with `starkli`


Farhad-Shabani:
## Context

To interact with IBC Cairo contracts, users can utilize `starkli` for declaring, deploying, and invoking contracts. While declaring and deploying IBC contracts with `starkli` is relatively straightforward, invoking them presents challenges due to the need for Cairo-encoded arguments, which hampers the user experience.

This issue is particularly relevant when wanted to submit a `MsgTransfer` to the ICS-20 Cairo contract. It involves constructing the message and serializing it into Cairo’s format. 

## Proposal
To address this, there should be a mechanism that either outputs a ready-to-use encoded value for users to pass directly into `starkli`, or alternatively, a command within the relayer’s CLI so that it handles the underlying complexity by accepting a few arguments like transfer amount and receiving address.

cc @soareschen, if it make sense to implement this as part of the relayer?

soareschen:
I am thinking that the simplest way is to build a simple CLI separately that would encode the messages with Cairo and output them as hex strings. The user can then copy and paste the string to starkli, or use bash subshells to compose them with the main starkli command.

This is simpler than providing it via the relayer, as a Starknet user won't need to figure out how to install and configure the relayer, together with the Cosmos config. On the other hand, we can reasonably expect all Starknet users know how to use starkli already.

rnbguy:
I wonder if it makes sense to look into some dApp frontends, such as
- [Scaffold-Stark](https://scaffoldstark.com)
- [Starknet-Scaffold](https://www.starknetscaffold.xyz)

rnbguy:
> simplest way is to build a simple CLI separately that would encode the messages with Cairo

> as a Starknet user won't need to figure out how to install and configure the relayer

I think, we should just provide the CLI via relayer and make the relayer configuration optional for that subcommand.
It should be similar to `hermes tx ft-transfer ...` except for the aut-relaying part (as our validation will be auto-clearing the packets).

I suggest this as it is an interm solution until the Polaris frontend supports our implementation. I don't think, we need a separate binary for this.

PR #232, opened 2025-01-27
Add CLI to generate `starkli` arguments for `send_transfer`

ljoss17:
Closes: #206 

This PR adds a CLI which takes the following arguments:

* `amount`
* `receiver`
* `port_id`
* `channel_id`
* `timeout-timestamp`
* `denom`

It will then create a `MsgTransfer` and encode it to a `Vec<Felt>`.
The output of the command is the values of the `Vec<Felt>` which can then be passed to the `starkli invoke <contract addr> send_transfer <args>`


relayer/crates/starknet-cli/src/commands/demo/transfer.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[clap(
+        long = "amount",
+        required = true,
+        value_name = "AMOUNT",
+        help_heading = "REQUIRED"
+    )]
+    pub amount: String,
+
+    /// Denom of the amount to transfer. If not specified this
+    /// will default to `stake`
```
rnbguy, line 40: Why default to `stake`? In starknet the tokens are erc20 contract address. This has to be a `Felt`.

PS. if you want to keep it optional, you can use the canonical `strk` token address -- this is known and fixed after a network is launched. So you may keep a `const` that is passed here.

You can use [`std::env`](https://doc.rust-lang.org/std/macro.env.html) macro to pass the value at compile time.

relayer/crates/starknet-cli/src/commands/demo/transfer.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[clap(long = "denom", value_name = "DENOM")]
+    pub denom: Option<String>,
```
rnbguy, line 42: I think, it's best not to make it optional.

ljoss17: Done in c24daa9e615f165c26eadd32760063b257f2ed64

relayer/crates/starknet-cli/src/commands/demo/transfer.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub amount: String,
+
+    /// Denom of the amount to transfer
+    #[clap(
+        long = "denom",
+        required = true,
+        value_name = "DENOM",
+        help_heading = "REQUIRED"
+    )]
+    pub denom: String,
```
rnbguy, line 46: add a flag like `--erc20` that makes sure the passed denom is erc20 contract address (felt) and keep the denom prefix empty.

rnbguy: you can also make the denom an `enum` and handle is appropriately.

tools/crates/cli/src/contexts/app.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        let relayer_wallet = chain_driver
+            .wallets
+            .get("relayer")
+            .ok_or_else(|| StarknetChainDriver::raise_error("expect relayer wallet to be present"))?
+            .clone();
+
+        let chain_config = StarknetChainConfig {
+            json_rpc_url: format!("http://localhost:{}/", chain_driver.node_config.rpc_port),
```
rnbguy, line 156: This should be [SocketAddr](https://doc.rust-lang.org/std/net/enum.SocketAddr.html). The public endpoint will have an IP.

ljoss17: Done in 3a61fb1907611e90e9923218b230f6638df1fb49

ljoss17: Will add the configurable IP in another commit

ljoss17: Added a `rpc_addr` field to the `StarknetNodeConfig` but the bootstrap currently uses localhost and a random free port. See https://github.com/informalsystems/ibc-starknet/pull/232/files#diff-e7746581e0a2ba08412a5d6307cbe2118ef587b5ee2258a5b37534ed0df0c531R62

relayer/crates/starknet-cli/src/commands/demo/transfer.rs
ljoss17: Done in 80af277cfe504533593990f2d29b4095cbc417f8

tools/crates/cli/src/commands/starknet/transfer_args.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub struct RunTransferArgs;
+
+#[async_trait]
+impl CommandRunner<ToolApp, TransferArgs> for RunTransferArgs {
+    async fn run_command(
+        app: &ToolApp,
+        args: &TransferArgs,
+    ) -> Result<<ToolApp as HasOutputType>::Output, <ToolApp as HasErrorType>::Error> {
+        let builder = app.load_builder().await?;
```
Farhad-Shabani, line 80: This line attempts to load a `config.toml` file while loading the app, and it's not clear where that file should live. If that's necessary for this CLI tool, I suggest the CLI either be taking the file as an argument or have the required parameters hardcoded for better independent control over them.

Farhad-Shabani (commented):
Great job @ljoss17!

Left a few comments. Hope they make sense! 
Additionally, it would be great to have a very brief README outlining how this binary works and how the returned value should be consumed by users. Like a list of needed files and  commands to execute a successful send transfer to the testnet. Thanks.

tools/crates/cli/src/commands/starknet/transfer_args.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub receiver: String,
+
+    /// Address of the sender
+    #[clap(
+        long = "sender",
+        required = true,
+        value_name = "SENDER",
+        help_heading = "REQUIRED"
+    )]
+    pub sender: String,
```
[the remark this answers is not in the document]
ljoss17: Done in 9c4651dd614520ac8a178101975a5690c309810e

tools/crates/cli/src/commands/starknet/subcommand.rs
```diff
@@ -0,0 +1,29 @@
+use cgp::core::Async;
+use hermes_cli_components::traits::command::{CanRunCommand, CommandRunner};
+
+use crate::commands::starknet::transfer_args::TransferArgs;
+
+#[derive(Debug, clap::Subcommand)]
+pub enum StarknetSubCommand {
+    TransferArgs(TransferArgs),
```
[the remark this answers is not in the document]
ljoss17: Done in e84c218caf1808477a99c3c1529ccddff688bf00

tools/crates/cli/src/contexts/app.rs
```diff
@@ -0,0 +1,186 @@
...
+{
+    type AnyCounterparty = CosmosChain;
+}
+
+impl<Value> OutputProducer<ToolApp, Value> for ToolAppComponents {
+    fn produce_output(_app: &ToolApp, _value: Value) {}
+}
+
+impl ConfigUpdater<StarknetChainDriver, StarknetRelayerConfig> for UpdateToolConfig {
+    fn update_config(
```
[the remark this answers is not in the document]
ljoss17: Not required, removed in fc8a7c0bba0df940c149afb8d4d42a7ab37c7767

tools/crates/cli/src/commands/starknet/transfer_args.rs
ljoss17: Instead of using the chain_status to set the timeout, the user can now specify the explicit timeout. This allows us to remove the need to load the builder and other dependencies which made the UX worse. Done here 96ff0dafb62b61622c2ffdc26ff71f90c0caf234

relayer/crates/starknet-test-components/src/impls/bootstrap/bootstrap_chain.rs
```diff
@@ -58,6 +59,10 @@ where
             .await
             .map_err(Bootstrap::raise_error)?;
 
+        // FIXME: RPC address is set to localhost and port is set to a random free port
+        // The values should be configurable to connect to a specific node
```
soareschen, line 63: For bootstrapping, the host should always be localhost since we are running the chain process. If we are moving to testnet, we will just use a dummy bootstrap context that returns a chain with preconfigured address.

Farhad-Shabani (approved):

tools/crates/cli/src/contexts/app.rs
rnbguy: This should have been `String` or `Url` because of #289. You were right. Sorry for my oversight.

PR #242, opened 2025-01-31
feat(lc/starknet): verify signed commitment proofs from starknet

### LINKED ISSUE #229: [lc/starknet] verify membership methods don't have access to CosmWasm context
To avoid wasm blob size bloating, we are using cosmwasm's secp256k1 verifier as part of their native environment import. We successfully added this for client update in #205.

Now, similar to #228, we need to validate the commitment proofs too.

But to verify this commitment proof inside the cosmwasm light client, we implement them on [`ClientStateCommon`](https://docs.rs/ibc/latest/ibc/core/client/context/client_state/trait.ClientStateCommon.html) -- but this trait doesn't pass the [`ClientValidationContext`](https://docs.rs/ibc/latest/ibc/core/client/context/trait.ClientValidationContext.html). So, we can't really use the cosmwasm's native secp256k1 verifier.

Currently, we are considering forking https://github.com/informalsystems/cosmwasm-ibc and adding the signature verifier calls directly at [`Context` implementation](https://docs.rs/ibc-client-cw/latest/src/ibc_client_cw/handlers.rs.html#66-106).

PS. this issue is redundant if #40 is resolved.

rnbguy:
closes #229


relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/connection_message.rs
```diff
@@ -126,7 +158,7 @@ where
             },
             delay_period,
             update_height,
-            proof_init: counterparty_payload.proof_init.proof_bytes,
+            proof_init,
```
rnbguy, line 161: I successfully managed to sign the `proof_init`

relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/connection_message.rs
```diff
@@ -214,7 +246,8 @@ where
         let message = CosmosConnectionOpenConfirmMessage {
             connection_id: connection_id.to_string(),
             update_height,
-            proof_ack: counterparty_payload.proof_ack.proof_bytes,
+            // FIXME(rano): generate the signature
+            proof_ack: dbg!(counterparty_payload.proof_ack.proof_bytes),
```
rnbguy, line 250: Currently, the signature verification fails here. But I don't know how to create the signature at this point without access to `ConnectionEnd` and `ConnectionId` at starknet.

rnbguy: I can query the stored counteryparty connection end (starknet) at cosmos, cosmos's own connection id to recreate the ConnectionEnd.

rnbguy:
To generate the proofs -- specifically signatures, I hacked around. This is not ideal.

To generate a signature, I need
- starknet consensus root at cosmos (against which the proof is validated)
- starknet commitment prefix
- ibc path
- ibc value

When I am building the message with CosmosChain, I don't have access to StarknetChain. So I can't fetch the consensus_root (block_hash) for starknet.

For now, I pass the `block_hash` as a dummy proof in the payload -- as I am building the correct proof at message builder anyway.

@soareschen any idea how to do this properly?

soareschen:
I think you need to include that in the payload. You can define a custom payload type for Starknet if needed.

soareschen:
You could also modify the `StarknetCommitmentProof` type if it is required in all proofs. I think in Cosmos, the root is part of the `MerkleProof` struct already.

rnbguy:
I ended up building the bytes that need to be signed at the Starknet (-to-Cosmos) payload builder step. But Starknet doesn't have access to the Cosmos's default signer. So I signed it at Cosmos's message builder step.

But I don't see how I can sign the bytes for packet commitment, ack and receipt.

rnbguy:
One more question: is the `query_packet_receipt` supposed to return `non_membership_proof` too?

https://github.com/informalsystems/ibc-starknet/blob/a3248f9d7c1c352f155fc7f17e2df92fd220a048/relayer/crates/starknet-chain-components/src/impls/queries/packet_receipt.rs#L51-L57

rnbguy:
currently fails because the `packet_ack` proof from Starknet to Cosmos is not valid, i.e. it is not signed by the cosmos default signer.

rnbguy:
Actually, I think, the `default_message_signer` (or something equivalent) should be used on the Starknet side at payload builder steps.

- It only makes sense as the public key is part of the Starknet light client. So, it should be from Starknet.
- We can generate the proper commitment proofs at payload builder steps directly. Proofs from Starknet should only rely on Starknet's configuration.

Starknet does have an equivalent to `default_message_signer` via `HasStarknetAccount`

https://github.com/informalsystems/ibc-starknet/blob/c14166fe1b862056b191a3909184adc8caf08979/relayer/crates/starknet-chain-context/src/contexts/chain.rs#L192

But, it uses [`ecdsa` scheme](https://docs.rs/starknet/latest/starknet/signers/trait.Signer.html#tymethod.sign_hash).

rnbguy:
I fixed the signing problems. But now I think, I need to make the commitment proofs consistent between ibc-go and cairo contract.

This is how `ibc-go` generate the commitment proof

https://github.com/cosmos/ibc-go/blob/98d7e7550a23ecf8d96ce042ab11ef857b184f2a/modules/core/04-channel/types/packet.go#L20-L36

In cairo, we use the following

https://github.com/informalsystems/ibc-starknet/blob/7967c8045ed6b4453030e01d0df12c47c2d77b37/cairo-contracts/packages/core/src/commitment/types.cairo#L51-L59

I should try to fix it using this test as a reference,

https://github.com/informalsystems/ibc-starknet/blob/7967c8045ed6b4453030e01d0df12c47c2d77b37/cairo-contracts/packages/core/src/tests/commitment.cairo#L52-L58

relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
```diff
@@ -84,27 +91,54 @@ where
...
-        // FIXME: properly fetch consensus state with proofs
+        // FIXME: properly fetch client state with proofs
         let client_state = chain
             .query_client_state(tag, client_id, query_height)
             .await?;
 
+        let chain_status = chain.query_chain_status().await?;
+
+        // FIXME: CometClientState can't be encoded to protobuf
+        let protobuf_encoded_client_state = Vec::new();
```
rnbguy, line 119: do we have an implementation that encodes `CometClientState` to protobuf encoding?

cc @soareschen

PS. it's not important for demo2, as ibc-go currently ignored self client/consensus validation.

relayer/crates/starknet-chain-components/src/impls/queries/consensus_state.rs
```diff
@@ -128,9 +140,36 @@ where
             .query_consensus_state(tag, client_id, consensus_height, query_height)
             .await?;
 
+        let chain_status = chain.query_chain_status().await?;
+
+        // FIXME: CometConsensusState can't be encoded to protobuf
+        let protobuf_encoded_consensus_state = Vec::new();
```
rnbguy, line 146: @soareschen same as client state, but for consensus state.

rnbguy:
We have to remove or trim the `light_client.rs` tesfile.

We are passing dummy proofs here:

https://github.com/informalsystems/ibc-starknet/blob/7967c8045ed6b4453030e01d0df12c47c2d77b37/relayer/crates/starknet-integration-tests/src/tests/light_client.rs#L394-L396

https://github.com/informalsystems/ibc-starknet/blob/7967c8045ed6b4453030e01d0df12c47c2d77b37/relayer/crates/starknet-integration-tests/src/tests/light_client.rs#L442

The wasm light client accepts only signed proofs now.

relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
soareschen: We don't have protobuf encoding for `CometClientState`. But what we should probably do is to use back `TendermintClientState` and implement Cairo encoding for that. We currently have `CometClientState` because the fields don't exactly match.

relayer/crates/starknet-chain-components/src/traits/proof_signer.rs
```diff
@@ -0,0 +1,29 @@
...
+#[cgp_component {
+  name: StarknetProofSignerTypeComponent,
+  provider: ProvideStarknetProofSignerType,
+  context: Chain,
+}]
+pub trait HasStarknetProofSignerType: Async {
+    type ProofSigner: Async;
+}
+
+#[cgp_component {
```
soareschen, line 12: With new version of `cgp`, we can now use `cgp_getter` here, and then you get an implementation of `UseField<FieldName>` and `UseFields`, which uses the default `proof_signer` field name. That way, you don't need to manually implement `GetStarknetProofSignerField`.

soareschen (approved):

relayer/crates/starknet-chain-components/src/traits/proof_signer.rs
rnbguy: ah I thought, this is recommended as I saw it in the code. 😅

Thanks for catching this ! I will use `cgp_getter`.

rnbguy: Looks like, I am seeing some other error when I try to use `cgp_auto_getter`.

```
conflicting implementations of trait `hermes_encoding_components::traits::has_encoding::ProvideEncodingType<contexts::chain::StarknetChain, hermes_cairo_encoding_components::types::as_felt::AsFelt>` for type `contexts::chain::StarknetChainContextComponents`
conflicting implementation in crate `hermes_encoding_components`:
- impl<Component, Context, Kind> hermes_encoding_components::traits::has_encoding::ProvideEncodingType<Context, Kind> for Component
  where Context: cgp::prelude::Async, Component: cgp::prelude::DelegateComponent<hermes_encoding_components::traits::has_encoding::EncodingTypeComponent>, <Component as cgp::prelude::DelegateComponent<hermes_encoding_components::traits::has_encoding::EncodingTypeComponent>>::Delegate: hermes_encoding_components::traits::has_encoding::ProvideEncodingType<Context, Kind>;
```

I will go ahead and merge it, but I will open an issue about using `cgp_auto_getter`.

soareschen: There are two macros. `#[cgp_auto_getter]` always require the context to have the field with the same name, and does not require explicit wiring. On ther other hand, `#[cgp_getter]` requires explicit wiring, but provides default providers like `UseFields` which means there is no need for manual implementation most of the time.

More info at: https://patterns.contextgeneric.dev/generic-accessor-providers.html

PR #244, opened 2025-01-31
fix(relayer): use starknet block timestamp

### LINKED ISSUE #243: [relayer] avoid using `Timestamp::now()` or `Time::now()`
many of the starknet components uses the relayer node timestamp as starknet time.

we should rather pick the timestamp from [`BlockWithTxHashes`](https://docs.rs/starknet/latest/starknet/core/types/struct.BlockWithTxHashes.html).

rnbguy:
closes #243

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
@@ -37,6 +37,7 @@ where
         + HasBlobType<Blob = Vec<Felt>>
         + HasEncoding<AsFelt, Encoding = Encoding>
         + CanQueryContractAddress<symbol!("ibc_client_contract_address")>
+        + CanRaiseAsyncError<&'static str>
         + CanRaiseAsyncError<Encoding::Error>,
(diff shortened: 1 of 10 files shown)


relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
```diff
@@ -57,10 +58,10 @@ where
         let client_id_seq = client_id
             .as_str()
             .rsplit_once('-')
-            .expect("valid client id")
+            .ok_or_else(|| Chain::raise_error("valid client id"))?
             .1
             .parse::<u64>()
-            .expect("valid sequence");
+            .map_err(|_| Chain::raise_error("valid sequence"))?;
```
soareschen, line 64: The error should probably be "invalid" instead of "valid". We could also use `CanRaiseError<ParseIntError>` for this case.

soareschen (approved):

relayer/crates/starknet-chain-components/src/impls/queries/client_state.rs
rnbguy: nice catch. fixed at feb5c16

Issue #220, opened 2025-01-21
Implement chain methods to support packet clearing on relayer


soareschen:
- `CanQueryPacketCommitments`
  - `CanQueryPacketAcknowledgements`
  - `CanQuerySendPackets`
  - `CanQueryUnreceivedPacketSequences`
  - `CanQueryUnreceivedAcksSequences`

soareschen:
This has been superceded by https://github.com/informalsystems/hermes-sdk/pull/535

PR #250, opened 2025-02-04
Implement `UnreceivedPacketSequencesQuerier` for StarknetChain

ljoss17:
This PR adds the `CanQueryUnreceivedPacketSequences` implementation required for packet clearing.

See https://github.com/informalsystems/ibc-starknet/issues/220


relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -0,0 +1,599 @@
...
+        let balance_cosmos_a_step_0 = cosmos_chain
+            .query_balance(address_cosmos_a, denom_cosmos)
+            .await?;
+
+        info!(
+            "cosmos balance before transfer: {}",
+            balance_cosmos_a_step_0
+        );
+
+        let commitment_sequences = vec!(Sequence::from(1));
```
rnbguy, line 402: why did you write it as `vec!()` 😅

relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -0,0 +1,599 @@
...
+            &cosmos_channel_id,
+            &IbcPortId::transfer(),
+            wallet_cosmos_a,
+            address_starknet_b,
+            &Amount::new(transfer_quantity, denom_cosmos.clone()),
+            &None,
+        )
+        .await?;
+
+        assert_eq!(pending_before, commitment_sequences);
```
rnbguy, line 421: We can't add these asserts in `ics20.rs` as we are using auto-relaying, right?

ljoss17: Exactly, and as mentioned at the top, this test will undergo changes when adding the query pending acks. As this will require doing partial relaying manually.

rnbguy (approved):
Thanks, Luca ! 🙌🏽

Issue #270, opened 2025-02-10
incorrect channel and port id in channel open try handler


rnbguy:
https://github.com/informalsystems/ibc-starknet/blob/d751ba2ebb21b31d472b48171844c2afb1a9d374/cairo-contracts/packages/core/src/channel/components/handler.cairo#L122

should be

```rs
                .read_channel_end(@msg.packet.port_id_on_a, @msg.packet.chan_id_on_a);
```

---

https://github.com/informalsystems/ibc-starknet/blob/d751ba2ebb21b31d472b48171844c2afb1a9d374/cairo-contracts/packages/core/src/channel/components/handler.cairo#L757

should be

```rs
            chan_end_on_a.validate(packet.port_id_on_b, packet.chan_id_on_b);
```

---

PS. These kinds of bugs will be difficult to find out if we start from a fresh setup. I found these bugs by initializing Cairo contract's `next_client_sequence`, `next_connection_sequence`, `next_channel_sequence` with random non-zero value -- to generate different identifiers from the Cosmos chain.

PR #271, opened 2025-02-10
fix: use correct port and channel IDs during channel end retrieval in `ack_packet`

Farhad-Shabani:
Closes: #270

### DIFF
--- a/cairo-contracts/packages/apps/src/tests/transfer.cairo
+++ b/cairo-contracts/packages/apps/src/tests/transfer.cairo
@@ -121,7 +121,7 @@ fn test_unescrow_ok() {
 
     let prefixed_denom = cfg.prefix_native_denom();
 
-    let recv_packet = cfg.dummy_packet(prefixed_denom.clone(), COSMOS(), STARKNET());
+    let recv_packet = cfg.dummy_incoming_packet(prefixed_denom.clone(), COSMOS(), STARKNET());
 
     // Submit a `RecvPacket` to the `TransferApp` contract.
     ics20.on_recv_packet(recv_packet);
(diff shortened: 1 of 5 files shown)


rnbguy:
Thanks for the quick fix. I will merge it tomorrow.

I realized, if we initialize the identifier sequence counter with random values -- we can catch these inconsistencies. Do you think, you can add a few tests with this idea in your cairo integration tests?

Farhad-Shabani:
> Thanks for the quick fix. I will merge it tomorrow.

Of course!

> I realized, if we initialize the identifier sequence counter with random values -- we can catch these inconsistencies. Do you think, you can add a few tests with this idea in your cairo integration tests?

About the identifier sequences on Starkent, because of the logic in place - they are only incremented once a client, connection, or channel created - we can't simply go with random numbers, since it would fail bunch of other existing tests.  That said, I totally see what you'd like to achieve. I think, ideally, these random numbers should be generated for the IDs of counterparty chains, so won't be having the same IDs as on Starknet. However, I couldn’t find any random number generator in Cairo that we could use in tests. (There is a potential approach, it isn't applicable in a testing environment).

Anyhow, I have made improvements to some semantics related to dummy packet creation. Specifically, I have differentiated between incoming and outgoing packets and set the counterparty channel ID to a sufficiently high value to avoid conflicts with existing IDs on Starknet. I believe this effectively should cover the relevant scenarios for now. See please if it works with you?

rnbguy:
> with random numbers

I misused the term 😅 I meant arbitrary different constants for identifiers between two hosts.

You totally got my point. The use of `10` in your tests is exactly what I wanted. Thanks ! 🙌🏼

rnbguy (approved):

Issue #277, opened 2025-02-11
Implement Cairo contract query endpoints for send/ack commitment sequencess


soareschen:
Part of: https://github.com/informalsystems/ibc-starknet/issues/218

To support packet clearing, the relayer needs to query for a list of sequences which IBC Starknet has committed for Send/Ack, so that it can use them on the counterparty chain to find out which packet has not yet been relayed.

https://github.com/informalsystems/ibc-starknet/pull/226 has implemented the unreceived send/ack query methods, but those are for processing the sequences coming from Cosmos. We would need the opposite query methods for clearing in the other direction.

Following are the type signatures for reference:

```rust
pub trait CanQueryPacketCommitments<Counterparty>:
    HasHeightType
    + HasChannelIdType<Counterparty>
    + HasPortIdType<Counterparty>
    + HasSequenceType<Counterparty>
    + HasAsyncErrorType
{
    async fn query_packet_commitments(
        &self,
        channel_id: &Self::ChannelId,
        port_id: &Self::PortId,
    ) -> Result<Vec<Self::Sequence>, Self::Error>;
}

pub trait CanQueryPacketAcknowledgements<Counterparty>:
    HasHeightType + HasChannelIdType<Counterparty> + HasPortIdType<Counterparty> + HasAsyncErrorType
where
    Counterparty: HasSequenceType<Self>,
{
    async fn query_packet_acknowlegements(
        &self,
        channel_id: &Self::ChannelId,
        port_id: &Self::PortId,
        sequences: &[Counterparty::Sequence],
    ) -> Result<Option<Vec<Counterparty::Sequence>>, Self::Error>;
}
```

PR #278, opened 2025-02-11
feat: implement query send/ack packet sequences endpoints

Farhad-Shabani:
Closes: #277

### DIFF
--- a/cairo-contracts/packages/core/src/channel/components/handler.cairo
+++ b/cairo-contracts/packages/core/src/channel/components/handler.cairo
@@ -174,6 +174,39 @@ pub mod ChannelHandlerComponent {
             self.read_packet_ack(@port_id, @channel_id, @sequence)
         }
 
+        fn packet_commitment_sequences(
+            self: @ComponentState<TContractState>, port_id: PortId, channel_id: ChannelId
+        ) -> Array<Sequence> {
+            let mut send_sequences = ArrayTrait::new();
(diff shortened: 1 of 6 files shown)


cairo-contracts/packages/core/src/channel/components/handler.cairo
```diff
@@ -174,6 +174,39 @@ pub mod ChannelHandlerComponent {
             self.read_packet_ack(@port_id, @channel_id, @sequence)
         }
 
+        fn packet_commitment_sequences(
```
Farhad-Shabani, line 177: @soareschen, this is based on the logic we chatted. 
Just one thing—when the number of packet sequences gets really large, say 100,000, it's going to iterate through storage a massive number of times, which doesn’t seem practical. 
Maybe we should set a limit? Do you have any suggestions for the number of iterations that suffice for the relayer?

soareschen: Yes, I think one way is that we have a pagination limitation of something like 1000. It may also be good if the function accepts an optional `latest_sequence` parameter, so that the relayer may try to query further into the past in case if needed.

In any case, I think packet clearing is mainly used for recently missed packets. So we probably don't need to go back too many sequences. We could actually also go through the event-based relaying by polling for events in the past, instead of the present.

soareschen (approved):

Issue #279, opened 2025-02-12
ibc token naming at starknet


rnbguy:
We are using

https://github.com/informalsystems/ibc-starknet/blob/919216e8cb94093acc6da2ae710782594a4e6ed5/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L762-L764

So, when we are sending `OSMO` token -- which is represented as `uosmo` as base token at Osmosis, we create a ERC20 token with symbol `IBC/uosmo`. Ideally, this should have been `IBC/osmo`.

Farhad-Shabani:
@rnbguy, are you saying the `denom` in `MsgTransfer` should be set to `osmo`? Otherwise, if the contract is supposed to handle this, I’m a bit concerned. It would add costly logic to the ICS-20 contract. I am wondering how would the contract even know the exact token name? Like in the case of `uunity`, or if, in the future, Starknet connects to non-Cosmos chains?
Wouldn’t it make more sense for web3 frontend applications to handle this instead? So that, the user-facing side stays cleaner.

rnbguy:
Totally agree with your comment. I also don't know how to resolve this. Let's discuss during next week's sync.

rnbguy:
I came across a scenario, where we try to send an IBC token minted by a different chain on Osmosis to Starknet. In that case, the symbol would be `IBC/ibc/...` 😅

rnbguy:
With #285, we don't really need this anymore. By setting the denom exponent (ERC20 token decimal) to zero, we are explicitly using a base denom unit. In that case, `IBC/uosmo` is the correct choice for symbol.

rnbguy:
closed by #285 

we are using decimal zero now. So using base denom here is technically correct.

Note, ideally we want to use the original denom and decimals. moving that discussion to new issue.

PR #285, opened 2025-02-17
fix: set ERC20 token decimal to zero via `ERC20Metadata` impl

### LINKED ISSUE #274: consistent denom exponent
```
1 ERC20 = 10^18 BASE_ERC20
```
but
```
1 COSMOS_DENOM = 10^6 BASE COSMOS_DENOM
```

For now, we mint/burn `1 BASE_ERC20` for `1 BASE COSMOS_DENOM`. But that makes `1 ERC20 = 10^12 COSMOS_DENOM`.

This can lead to UX issues. We should have `1 ERC20` mint/burn for `1 COSMOS_DENOM`.

ref: https://docs.cosmos.network/main/build/architecture/adr-024-coin-metadata

rnbguy:
closes #274


rnbguy:
Uses [the static approach](https://docs.openzeppelin.com/contracts-cairo/0.20.0/erc20#the_static_approach) from OZ docs. We can decide between this static approach or [the storage approach](https://docs.openzeppelin.com/contracts-cairo/0.20.0/erc20#the_storage_approach). 

How the token gets displayed on Starknet explorers now:
- `main` branch: [transaction](https://sepolia.voyager.online/tx/0x299d54bfff0f3f48755efe76aa65ca351c63034c17f053a7e5aaffab1f355b6), [metadata](https://sepolia.voyager.online/token/0x029b7f914e21cefad24aef61a94623a68dbb2e51c3212721f61bf11eed6620cc)
- this PR branch: [transaction](https://sepolia.voyager.online/tx/0x6c2e3886beef525e21de08ac647a7ed531d16551a5ed96009222fbfdb767cbd), [metadata](https://sepolia.voyager.online/token/0x059ef3c6c39333036d10f6228469cad98ec2ae3c39e2d41b73cfddc4eeb60b8d)

cairo-contracts/packages/contracts/src/erc20.cairo
```diff
@@ -19,9 +19,10 @@ pub mod ERC20Mintable {
     #[abi(embed_v0)]
     impl ERC20MintableImpl = ERC20MintableComponent::ERC20Mintable<ContractState>;
 
-    // ERC20 Mixin
     #[abi(embed_v0)]
-    impl ERC20MixinImpl = ERC20Component::ERC20MixinImpl<ContractState>;
```
rnbguy, line 24: From OZ doc,

> Note that we are not using the MixinImpl in this case, since we need to customize the IERC20Metadata implementation.

cairo-contracts/packages/contracts/src/erc20.cairo
```diff
@@ -59,4 +60,20 @@ pub mod ERC20Mintable {
...
+        fn name(self: @ContractState) -> ByteArray {
+            self.erc20.name()
+        }
+
+        fn symbol(self: @ContractState) -> ByteArray {
+            self.erc20.symbol()
+        }
+
+        fn decimals(self: @ContractState) -> u8 {
+            0
```
Farhad-Shabani, line 76: First off, my understanding is that smart contracts mostly (if not all) avoid floating-point arithmetic due to security risks and computational overhead. Instead, they operate with the smallest denom of a token  (similar to `wei` for `ETH` or `uosmo` for `OSMO`). Therefore this decimal setting is primarily used for user-facing applications, ensuring that the smallest unit is **correctly interpreted as whole tokens in UI/UX**. 

It sounds like on **Cosmos chains** a zero exponent tells the chain that the application **does not recognize any decimal places**, rather than explicitly saying the decimal is zero. If this is a semantic convention there, that’s fine then. Though, on Starknet, setting the decimal to zero can still lead to misinterpretations in UI.
Say if I transfer 5 `OSMO`, with a zero decimal, it would arrive on Starknet as 5,000,000 `IBC/uosmo`. While this is technically correct at the contract level, as a user, I would expect to see 5 `IBC/OSMO`. However, Voyager or any similar frontends wouldn’t be able to display it correctly since the decimal has been set to 0, leaving them without any context to display it properly.

That said, if non-zero decimals? then the storage approach sounds the correct way forward to me. Decimal handling isn’t uniform even within Cosmos-based chains. For example, Injective and Evmos use 18, while others use 6. On top of that, we should have an eye on the potential IBC between SN<>SN or SN<>other non-Cosmos chains.

All in all, I’m okay with zero as a temporary fix for the testnet to get things technically right, but I don’t think it really solves the underlying issue.

rnbguy: Have you read this comment at issue?

https://github.com/informalsystems/ibc-starknet/issues/274#issuecomment-2662909735

rnbguy: also, by suggesting 'IBC/OSMO' instead of 'IBC/uosmo', are you agreeing with my token denom issue?

Farhad-Shabani: > also, by suggesting 'IBC/OSMO' instead of 'IBC/uosmo', are you agreeing with my token denom issue #279?

Yup. In fact I'm getting more in favor of having the token creation step done by governance. This would ensure that the naming and decimal are set appropriately. Though take it as one possible solution, might not be the best as it's not convention in IBC world.

rnbguy: just to clarify, so you don't agree how things are happening at Cosmos side. is this right?

I agree with the above. But in that case, I don't know how else to fix this now other than a spec lebel change to pass denom metadata from cosmos.

Note, not all tokens at osmosis have exponent 6. the axelar-eth has exponent 18 (not part of osmosis Blockchain data, maintained by explorer or wallet) and axelar-wbtc has exponent 8. And they can be transferred to starknet.

rnbguy: I am really not in favor of having permissioned token creation. 😅 but you can bring this up in the next sync.

Farhad-Shabani: > I am really not in favor of having permissioned token creation as it deviates from original ics20 spec. 😅 but you can bring this up in the next sync to Marius.

Honestly, we should think about potential solutions. I just threw out one mighty remedy.

Farhad-Shabani:
@rnbguy what do you think about utilizing the [`memo` field](https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#using-the-memo-field)? get the decimal from there? even the symbol?

cairo-contracts/packages/contracts/src/erc20.cairo
Farhad-Shabani: https://github.com/informalsystems/ibc-starknet/pull/285#issuecomment-2666747985

rnbguy: > I could just accept it as a convention in Cosmos chains.

I think we don't have any other choice because we are reusing the ibc-go's vanilla implementation. We have to accept, however, the ibc-go builds the Transfer packet data. 😅

> utilizing the memo field?

Can you please explain more? Note, ICS20 is fungible token transfer. So, people may send the same OSMO token with different metadata in the memo field. So, we will mint different tokens for the same cosmos denoms but with different metadata. IMO, this spoils the canonical IBC token on Starknet side.

rnbguy: I am thinking about memo field. I think, it is fine to utilize the memo field. It's the same UX as packet middleware. It's up to the user (or the wallet) to be mindful of what to put in the memo field.

If the user leaves the memo field empty, we fall back to what I have done in this PR.

I think, this is a broader discussion to be had with Osmosis -- especially the wallet who will support this.

For now, I am changing the PR for the `storage approach`. So, after the discussion, it'd be easy to transition.

cairo-contracts/packages/apps/src/transfer/erc20_call.cairo
```diff
@@ -42,17 +41,14 @@ pub impl ERC20ContractImpl of ERC20ContractTrait {
...
         owner: ContractAddress,
     ) -> ERC20Contract {
         let mut call_data = array![];
 
-        call_data.append_serde(name);
-        call_data.append_serde(symbol);
-        call_data.append_serde(amount);
-        call_data.append_serde(recipient);
-        call_data.append_serde(owner);
+        (name, symbol, decimals, amount, recipient, owner).serialize(ref call_data);
```
rnbguy, line 51: I refactored the code here like this, as I feel, semantically, this is correct.

Also, we don't need `SerializeAppend` anymore.

cairo-contracts/packages/contracts/src/erc20.cairo
Farhad-Shabani: > I am thinking about memo field. I think, it is fine to utilize the memo field. It's the same UX as packet middleware. It's up to the user (or the wallet) to be mindful of what to put in the memo field.

You got it! I was thinking the same way, and agree this should be discussed with the Osmosis team. Though, overall, the `memo` field option sounds to me the most promising way to go.

One possible refinement on top of this: 
We could require the `memo` to include the `symbol` (bare w/o `IBC/` since it's prepended by ICS20), and `decimal` only for the first token transfer (basically when the token is first created). After that, users should submit `MsgTransfer` with an empty memo. Since that info is already stored on-chain as metadata, there's no need to make users/apps include it every time. Just the first time.

```Cairo
struct MsgTransfer {
    pub port_id_on_a: PortId,
    pub chan_id_on_a: ChannelId,
    pub denom: PrefixedDenom,
    pub amount: u256,
    pub receiver: ByteArray,
    pub memo: Option<Memo>,
    pub timeout_height_on_b: Height,
    pub timeout_timestamp_on_b: Timestamp,
}
 
struct Memo {
    pub erc20_metadata: ERC20MetaData,
}

struct ERC20MetaData {
    pub symbol: ByteArray,
    pub decimal: u8,
}
```

cairo-contracts/packages/contracts/src/erc20.cairo
```diff
@@ -32,6 +34,9 @@ pub mod ERC20Mintable {
         mintable: ERC20MintableComponent::Storage,
         #[substorage(v0)]
         erc20: ERC20Component::Storage,
+        // The decimals value is stored locally in the contract.
```
Farhad-Shabani, line 37: Came across this update: https://github.com/OpenZeppelin/cairo-contracts/pull/1294
Hopefully, it'll be available soon in their next release, so we can go with that and won't need to keep this internal state anymore.

Farhad-Shabani (approved):
I approve. Appreciate your patience with my comments!
That said, this still feels like a temporary fix to me. I don’t see any blockers to going with the `memo` option, regardless of our discussion with the Osmosis team. That’s more of a coordination thing. But I’ll leave it up to you whether to upgrade things later, keep the issue open, or start a new one regarding https://github.com/informalsystems/ibc-starknet/pull/285#discussion_r1962389229, and https://github.com/informalsystems/ibc-starknet/pull/285#discussion_r1962410205.
Thanks Rano for all the investigations around.

cairo-contracts/packages/contracts/src/erc20.cairo
rnbguy: > decimal only for the first token transfer (basically when the token is first created)

It is not good from a permissionless perspective. As bad actor can try to snipe these token symbol creations. 😅

rnbguy: I also thought about DAO based token registration. I am fine if we do it after the token creation. As in, DAO (right now, it can be the relayer address) can look at a ICS20 token and update its symbol and decimals.

Like, right now, we are creating `IBC/uosmo` with zero decimal. Upon a (social) consensus, the DAO may update this to `OSMO` with 6 decimals for UX convenience.

That way, the token relaying stays permissionless. The popular tokens gets proper symbol and decimals via DAO.

cairo-contracts/packages/contracts/src/erc20.cairo
rnbguy: Yes I saw this too 👍🏼 but this is more in the `static approach`. The `storage approach` still remains the same.

rnbguy: I didn't open an issue for this, as the storage approach stayed the same.

PR #286, opened 2025-02-17
Update Hermes SDK, cgp, and scaffold packet clearing implementation

soareschen:
This PR tracks a number of changes from Hermes SDK and CGP, and prepare the ground work for a simplified support of packet clearing for Starknet.

Related PRs:
- https://github.com/informalsystems/hermes-sdk/pull/539
- https://github.com/informalsystems/hermes-sdk/pull/535
- https://github.com/informalsystems/hermes-sdk/pull/540
- https://github.com/informalsystems/hermes-sdk/pull/536


relayer/crates/starknet-chain-components/src/impls/queries/packet_is_cleared.rs
```diff
@@ -0,0 +1,25 @@
...
+        + HasSequenceType<Counterparty>
+        + HasAsyncErrorType,
+{
+    async fn query_packet_is_cleared(
+        _chain: &Chain,
+        _port_id: &Chain::PortId,
+        _channel_id: &Chain::ChannelId,
+        _sequence: &Chain::Sequence,
+    ) -> Result<bool, Chain::Error> {
+        Ok(false) // stub
```
rnbguy, line 23: why is this not implemented?

soareschen: I just haven't figured out how to implement this. It seems like we cannot use the `packet_commitment` call from the Cairo contract, as it would error if the commitment entry doesn't exist. But we cannot distinguish errors caused by the packet commitment not found, or by other reasons.

We probably need a separate endpoint to be able to query whether a packet commitment is found, or we could modify the existing endpoint to return an `Option`. @Farhad-Shabani how do you think?

In any case, this PR is mainly to scaffold the boilerplate for the new packet clearing, so we can fill in the proper implementation in another PR.

rnbguy: Ah I see.

Since this is specific to this implementation, let's wrap this PR and move this discussion to a new issue.

- Please open an issue for this.
- Please add a comment here with a brief description with the issue number.

Then you can merge this onto `main` 🙌🏼

soareschen: Filed https://github.com/informalsystems/ibc-starknet/issues/291

Farhad-Shabani: Opened https://github.com/informalsystems/ibc-starknet/pull/297

PR #287, opened 2025-02-17
Merge tools CLI into the relayer workspace

soareschen:
The `tools` project shares almost all dependencies as the relayer, and so should be put in the same Cargo workspace. When it is placed in a different workspace, it would require compiling all relayer dependencies twice, which slows down development.


rnbguy:
Perfect! Actually, I don't like that we have multiple cargo workspaces `light-clients` and `relayer`. Ideally, we should have one single workspace and be able to run every crate from the project root directory.

It is the same situation with Cairo -- we have two workspaces `cairo-libs` and `cairo-contracts`.

rnbguy (approved):
LGTM ! would be great to have all the workspaces unified in future.

soareschen:
For light client, it is slightly different, as there are much less dependencies needed for the light client. We may also want to move it to a separate repository when the project becomes more mature.

I think whether to place projects in the same workspace depends on how they affects Cargo.lock. When a subproject has much more dependencies than another subproject, a potential issue is that the locked dependencies may get "polluted" by the other external dependencies.

Anyway, we can probably also consider merging the light client workspace in a separate PR, if that helps with speeding up development.

PR #292, opened 2025-02-18
Apply `#[cgp_provider]` to all provider implementations

soareschen:
This follows the changes in https://github.com/informalsystems/hermes-sdk/pull/534 to apply `#[cgp_provider]` to all provider implementations. With this, we can now get much better error diagnostics by probling the error from `Context: CanUseComponent<ComponentName>`.

We also follow https://github.com/informalsystems/hermes-sdk/pull/541 to apply `#[cgp::re_export_imports]` to simplify the re-exporting of component constructs when defining presets.


rnbguy (approved):
Didn't do a thorough check. But, it looks ok.

soareschen:
It's fine to just roughly scan through. They are just boilerplate changes and don't involve actual change of logic.

PR #293, opened 2025-02-18
fix: use `max_clock_drift` in comet light client

### LINKED ISSUE #288: future timestamp is not allowed during cometbft light client update
https://github.com/informalsystems/ibc-starknet/blob/e0e617bfae7d9b155e1eded66c14682e81b46309/cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo#L34-L35

Probably we don't need this assert. We don't have this check at [ibc-rs](https://github.com/informalsystems/ibc-rs/blob/be82d123448a59ccea305c9e918f07aca5ec1a6f/ibc-clients/ics07-tendermint/src/client_state/common.rs#L174-L194) or [ibc-go](https://github.com/cosmos/ibc-go/blob/1f1110717d5d3a6c732149ec063e20a09b592e7f/modules/light-clients/07-tendermint/client_state.go#L79-L101).

Also, in testnet, we noticed that because of 30 sec block time, it's (almost) always `host_timestamp` is in the past.

rnbguy:
closes #288

### DIFF
--- a/cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
@@ -31,8 +31,23 @@ pub impl CometConsensusStateImpl of CometConsensusStateTrait {
         maybe_consensus_state.unwrap()
     }
 
-    fn status(self: @CometConsensusState, host_timestamp: u64, trusting_period: u64) -> Status {
-        assert(host_timestamp >= self.timestamp(), CometErrors::INVALID_HEADER_TIMESTAMP);
+    fn status(
+        self: @CometConsensusState, host_timestamp: u64, trusting_period: u64, max_clock_drift: u64,
+    ) -> Status {
+        // Consider `clock_drift` on Starknet with large `block_time`.
+        //
+        // `block_timestamp` is when the sequencer started building the block, which may be in the
+        // past.
+        // Without the `clock_drift`, the untrusted consensus state would be considered from future
+        // and fail.
+        //
+        // PS. the following check is always successful once a consensus state is validated and
+        // accepted as trusted.
+        // Because the `block_timestamp` is always increasing.
+        assert(
+            host_timestamp + max_clock_drift >= self.timestamp(),
+            CometErrors::HEADER_TIMESTAMP_FROM_FUTURE,
+        );
 
         let elapsed_time = host_timestamp - self.timestamp();
 
--- a/cairo-contracts/packages/clients/src/cometbft/errors.cairo
+++ b/cairo-contracts/packages/clients/src/cometbft/errors.cairo
@@ -4,7 +4,7 @@ pub mod CometErrors {
     pub const INVALID_CLIENT_STATE: felt252 = 'ICS07: invalid client state';
     pub const INVALID_CONSENSUS_STATE: felt252 = 'ICS07: invalid consensus state';
     pub const INVALID_HEADER: felt252 = 'ICS07: invalid header';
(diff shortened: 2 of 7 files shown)


cairo-contracts/packages/clients/src/cometbft/errors.cairo
```diff
@@ -4,7 +4,7 @@ pub mod CometErrors {
     pub const INVALID_CLIENT_STATE: felt252 = 'ICS07: invalid client state';
     pub const INVALID_CONSENSUS_STATE: felt252 = 'ICS07: invalid consensus state';
     pub const INVALID_HEADER: felt252 = 'ICS07: invalid header';
-    pub const INVALID_HEADER_TIMESTAMP: felt252 = 'ICS07: invalid header timestamp';
+    pub const HEADER_TIMESTAMP_FROM_FUTURE: felt252 = 'ICS07: header ts from future';
```
Farhad-Shabani, line 7: I’ve been trying to stick to an implicit convention for naming errors, kind of like what we did in `ibc-rs`, just to keep things consistent. Does the following make sense to you?

```suggestion
    pub const INVALID_HEADER_FROM_FUTURE: felt252 = 'ICS07: inv header from future';
```

cairo-contracts/packages/clients/src/tests/cometbft.cairo
```diff
@@ -46,25 +69,20 @@ fn test_client_sequence_ok() {
 
 #[test]
 fn test_update_client_ok() {
```
Farhad-Shabani, line 71: I noticed we're missing some other tests, like checking for client expiry or if it's frozen. Since we've been gradually adding new fields to the `ClientState`, it's possible now—like for max clock drift—where it wasn't before.
These tests can be built on top of the `simulate_update_client`, but I leave them for a separate PR.

cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
```diff
@@ -31,12 +31,31 @@ pub impl CometConsensusStateImpl of CometConsensusStateTrait {
...
+        let self_timestamp = self.timestamp();
 
-        if elapsed_time < trusting_period {
+        assert(
+            host_timestamp + max_clock_drift >= self_timestamp,
+            CometErrors::HEADER_TIMESTAMP_FROM_FUTURE,
+        );
+
+        // The header is in future but within the `max_clock_drift` period.
+        if host_timestamp < self_timestamp {
```
Farhad-Shabani, line 54: Added this condition, as for valid future headers this was underflowing.

rnbguy: Ah! Very good catch 👍🏼

cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
```diff
@@ -31,12 +31,31 @@ pub impl CometConsensusStateImpl of CometConsensusStateTrait {
...
+            return Status::Active;
+        }
+
+        if host_timestamp - self_timestamp < trusting_period {
```
rnbguy, line 58: but, maybe we merge the above condition as `host_timestamp < self_timestamp + trusting_period` here?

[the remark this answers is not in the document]
Farhad-Shabani: I honestly prefer that more explicit way over this compact one, especially since each test corresponds to one of the these conditions.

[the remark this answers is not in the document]
rnbguy: Ok. I don't mind. We can merge the PR if all is good.

Farhad-Shabani (approved):

PR #294, opened 2025-02-18
refactor: use `Url` for `json_rpc_url`

### LINKED ISSUE #289: starknet's `json_rpc_url` should be `Url`
https://github.com/informalsystems/ibc-starknet/blob/e0e617bfae7d9b155e1eded66c14682e81b46309/relayer/crates/starknet-chain-components/src/impls/types/config.rs#L15-L16

The public RPC endpoints are usually [a URL with domain names](https://docs.starknet.io/tools/fullnodes-rpc-providers/). We must make it `String` or `url::Url` to support them.

rnbguy:
closes #289

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/types/config.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/types/config.rs
@@ -1,5 +1,3 @@
-use std::net::SocketAddr;
-
 use hermes_cosmos_chain_components::impls::types::config::CosmosChainConfig;
 use serde::{Deserialize, Serialize};
 
@@ -13,6 +11,6 @@ pub struct StarknetRelayerConfig {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct StarknetChainConfig {
-    pub json_rpc_url: SocketAddr,
+    pub json_rpc_url: String,
     pub relayer_wallet: StarknetWallet,
 }
--- a/relayer/crates/starknet-cli/src/contexts/app.rs
+++ b/relayer/crates/starknet-cli/src/contexts/app.rs
@@ -1,4 +1,3 @@
-use std::net::SocketAddr;
 use std::path::PathBuf;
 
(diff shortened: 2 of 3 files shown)


relayer/crates/starknet-chain-components/src/impls/types/config.rs
```diff
@@ -13,6 +12,28 @@ pub struct StarknetRelayerConfig {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct StarknetChainConfig {
-    pub json_rpc_url: SocketAddr,
+    #[serde(with = "url_serde")]
```
soareschen, line 15: Don't the `url` crate implements `Serialize` for `Url` when the `serde` feature is enabled?

rnbguy: Just to clarify, this `Url` is from `starknet::providers::Url`.

soareschen: That is just a re-export from `url`.

rnbguy: Aah. I see 😅 I don't think, we can do much -- as the feature flag is not exposed via [`starknet` or `starknet-providers` feature flags](https://docs.rs/crate/starknet-providers/latest/source/Cargo.toml#36-37).

soareschen: We can import the `url` crate directly, and enable the feature there.

rnbguy: I don't want to do this as they are not using the latest version. So, we have to pin it to Starknet's version. Maybe we can just move to `String`, if that's ok with you.

rnbguy: Refactored using String: acd6759

soareschen (approved):

Issue #269, opened 2025-02-10
incorrect `connection_id` at `ChannelOpenTryMessage`


rnbguy:
`ChannelOpenTryMessage` for Cosmos must contain the expected channel end at `Cosmos`.

So the singleton `connection_hops` should be the connection ID at Cosmos, not at Starknet.

https://github.com/informalsystems/ibc-starknet/blob/d751ba2ebb21b31d472b48171844c2afb1a9d374/relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/channel_message.rs#L72-L79

To get the ConnectionID at Cosmos, we need the corresponding ConnectionEnd at Starknet -- which is not part of the `ChannelOpenTryPayload` at [hermes-sdk](https://github.com/informalsystems/hermes-sdk/blob/b432c9e2fa8d198d6294eb4e7567ed66efa23d7f/crates/chain/chain-components/src/types/payloads/channel.rs#L5-L12).

https://github.com/informalsystems/ibc-starknet/blob/d751ba2ebb21b31d472b48171844c2afb1a9d374/relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/channel_message.rs#L52

---

PS. These kinds of bugs will be difficult to find out if we start from a fresh setup. I found these bugs by initializing Cairo contract's `next_client_sequence`, `next_connection_sequence`, `next_channel_sequence` with random non-zero value -- to generate different identifiers from the Cosmos chain.

Issue #268, opened 2025-02-10
rm redundant message builders for cosmos


rnbguy:
We can reuse the message builders for cosmos directly from `hermes-sdk`.

The [channel_message](https://github.com/informalsystems/ibc-starknet/blob/7025c5a1ad17dd9b427007ed500f72a68ab492a8/relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/channel_message.rs) and [connection_message](https://github.com/informalsystems/ibc-starknet/blob/7025c5a1ad17dd9b427007ed500f72a68ab492a8/relayer/crates/starknet-chain-components/src/impls/starknet_to_cosmos/connection_message.rs) builders for starknet_to_cosmos are duplicating the same efforts from hermes-sdk.

If I am not mistaken, we can't reuse the implementation because StarknetChain has `u64` as `HeightType`. So we need to convert the height while building the Cosmos messages.

It would be great `hermes-sdk` adds a trait method for `HasHeightType` which implements the conversion to IBC Height type and reuses at these places.

PR #295, opened 2025-02-18
Use counterparty connection ID when constructing channel OpenTry message for Starknet

soareschen:
Closes: https://github.com/informalsystems/ibc-starknet/issues/269

Partially resolves the channel part of #268


relayer/crates/starknet-chain-components/src/impls/queries/counterparty_connection_id.rs
```diff
@@ -0,0 +1 @@
+
```
rnbguy, line 1: why there is an empty file? I think we remove this.

relayer/crates/starknet-chain-components/src/impls/queries/mod.rs
```diff
@@ -7,6 +7,7 @@ pub mod connection_end;
 pub mod consensus_state;
 pub mod contract_address;
 pub mod counterparty_chain_id;
+pub mod counterparty_connection_id;
```
rnbguy, line 10: and this is the corresponding `pub mod`.

relayer/crates/starknet-chain-components/src/components/starknet_to_cosmos.rs
```diff
@@ -54,6 +53,9 @@ mod preset {
                 CreateClientPayloadBuilderComponent,
                 UpdateClientPayloadBuilderComponent,
                 ChannelOpenInitMessageBuilderComponent,
+                ChannelOpenTryMessageBuilderComponent,
```
rnbguy, line 56: I think we can do this for Connection handshake too. Note, this closes #268.

relayer/crates/starknet-chain-components/src/impls/queries/counterparty_connection_id.rs
soareschen: Ah I wanted to implement the provider, but then figured we could reuse the same provider from Hermes SDK.

relayer/crates/starknet-chain-components/src/components/starknet_to_cosmos.rs
soareschen: I tried but it didn't work. Seems like there are some other conflicting constraints. We can fix that in separate PR.

rnbguy (approved):

Issue #276, opened 2025-02-11
Zero address appearing as sender in testnet when minting IBC tokens


Farhad-Shabani:
## Context
While testing a sample token transfer on testnet, it was observed that [the sender address appears as `0x00…0`](https://sepolia.starkscan.co/tx/0x0646abe322e8ed9b5b36ff41e6e75721480449ac749fc0f48fe0d2cb708996a8). This occurs because ICS-20 directly mints the token into the recipient’s account, making the sender address zero, instead of showing the actual ICS-20 contract as the sender.

## Proposal
We can first mint the tokens to the ICS-20 contract itself and then perform the transfer. This ensures that the ICS-20 address is correctly reflected as the sender. 

## Remark
~(Since token creation happens only once, this approach should not introduce any significant overhead.)~
It actually should be observed for the minting (not just token creation) process as well.

PR #275, opened 2025-02-11
imp: mint to ICS20 first then transfer to user

Farhad-Shabani:
Closes: #276

PR #296, opened 2025-02-18
refactor: transfer to ICS20 first then burn

Farhad-Shabani:
In continuation of #275
Closes: #276

### DIFF
--- a/relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
+++ b/relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
@@ -70,6 +70,7 @@ use ibc::core::host::types::identifiers::{PortId as IbcPortId, Sequence};
 use poseidon::Poseidon3Hasher;
 use sha2::{Digest, Sha256};
 use starknet::accounts::{Account, Call, ExecutionEncoding, SingleOwnerAccount};
+use starknet::core::types::U256;
 use starknet::macros::{selector, short_string};
 use starknet::providers::Provider;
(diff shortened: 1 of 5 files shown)


relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -542,6 +543,26 @@ fn test_query_unreceived_packets() -> Result<(), Error> {
...
+                selector: selector!("approve"),
+                calldata: call_data,
+            };
+
+            let execution = starknet_account_b.execute_v3(vec![call]);
+
+            let tx_hash = execution.send().await?.transaction_hash;
+
+            starknet_chain.poll_tx_response(&tx_hash).await?;
+        }
```
rnbguy, line 564: hmm. this is a very interesting ! So, the token owner can burn tokens from anyone's balance, but token owner can't transfer it to their own balance ?

Farhad-Shabani: Good question. This got me yesterday too. Should add a check in the `ERC20Mintable` contract.

In the standard ERC-20, since the supply is fixed, minting and burning don’t matter, so ownership isn’t really a concern. But in our case, we’re working with an extended version of ERC20 where only the owner (ICS20) can mint and burn tokens. This part of code was originally written at the start of the project just to get the MVP working. Now that it looks like we have no choice but to go with our own ERC20 version, it's good to review the implementation there and ensure it’s fully secure by adding some unit tests.

PR #298, opened 2025-02-19
Allow packet clearing to be done together with the `start` command

soareschen:
Closes: #248

Related: https://github.com/informalsystems/hermes-sdk/pull/542

This PR applies the patch in https://github.com/informalsystems/hermes-sdk/pull/542 to allow clearing heights to be passed to the `start` CLI to support packet clearing.

To slightly improve the UX, I have also split the Starknet `start` CLI into two subcommands: `hermes-starknet start starknet-with-cosmos` and `hermes-starknet start cosmos-with-starknet`. This mainly affects the positioning of chain A and B, so that the relayer uses the right chain to parse the chain IDs and other arguments.

In principle, we can make the CLI to automatically detect the chain type from the chain ID, and perform the appropriate dispatching. However, since we do not have the capacity to implement this feature, we can only use this quick hack to make the CLI slightly more position-independent.


ljoss17 (approved):
Looks good to me!

soareschen:
For `starknet-with-cosmos`, chain A is Starknet, while chain B is Cosmos, and vice versa. But maybe there is something wrong in the wiring some where. I will take a look later.

Issue #282, opened 2025-02-13
implement `oneof` in protobuf library


rnbguy:
We need to add [`oneof`](https://protobuf.dev/programming-guides/proto3/#oneof) support in our cairo protobuf library.

We specifically need it for two cases

- [`PublicKey`](https://github.com/cometbft/cometbft/blob/527f12dba30e29951d10815bf966ac6739342335/proto/tendermint/crypto/keys.proto#L11) in tendermint proto-defs
- [`CommitmentProof`](https://github.com/cosmos/ics23/blob/a324422529b8c00ead00b4dcee825867c494cddd/proto/cosmos/ics23/v1/proofs.proto#L90) in ics23 proto-defs

rnbguy:
The implementation of `PublicKey` in our current code should look something like this:

```cairo

#[derive(Default, Debug, Clone, Drop, PartialEq, Serde)]
pub enum PublicKey {
    #[default]
    Ed25519: ByteArray,
    Secp256k1: ByteArray,
}

impl PublicKeyAsProtoMessage of ProtoMessage<PublicKey> {
    fn encode_raw(self: @PublicKey, ref context: EncodeContext) {
        match self {
            PublicKey::Ed25519(ed25519) => {
                context.encode_field(1, ed25519);
            },
            PublicKey::Secp256k1(secp256k1) => {
                context.encode_field(2, secp256k1);
            },
        }
    }

    fn decode_raw(ref self: PublicKey, ref context: DecodeContext) {
        let tag = ProtobufTagImpl::decode(context.buffer[context.index]);

        if tag.field_number == 1 {
            let mut ed25519 = "";
            context.decode_field(1, ref ed25519);
            self = PublicKey::Ed25519(ed25519);
        } else if tag.field_number == 2 {
            let mut secp256k1 = "";
            context.decode_field(2, ref secp256k1);
            self = PublicKey::Secp256k1(secp256k1);
        } else {
            panic!("unexpected tag");
        }
    }

    fn wire_type() -> WireType {
        WireType::LengthDelimited
    }
}
```

We definitely should polish it in a PR.

PR #301, opened 2025-02-21
feat: implement Protobuf encoding for `Oneof` fields

Farhad-Shabani:
Closes: #282

## Highlight
A key change introduced in this PR is that `ProtoMessage` should only be implemented for structs that can serve as message types. Notably, `Oneof` and enums do not qualify as standalone message types, even though they are components of a protocol buffer message. To handle their (de)encoding, a separate trait definition and conversion mechanism is used, similar to the approach in the `prost` crate.


cairo-libs/packages/protobuf/src/tests/oneof.cairo
```diff
@@ -0,0 +1,127 @@
...
+                let n: u32 = e.into();
+                n.encode_raw(ref context);
+                let wire_type = ProtoMessage::<u32>::wire_type();
+                ProtobufTag { field_number: 3, wire_type }
+            },
+        }
+    }
+
+    fn decode_raw(ref context: DecodeContext, tag: ProtobufTag) -> Option<Oneof> {
+        if tag.field_number == 1 {
```
Farhad-Shabani, line 60: We couldn't use a `match` statement because it expects patterns to start from 0. Additionally, Cairo does not support `match` statements on tuples like (`field_number`, `wire_type`).

cairo-libs/packages/protobuf/src/types/message.cairo
```diff
@@ -63,6 +69,25 @@ pub impl EncodeContextImpl of EncodeContextTrait {
             i += 1;
         }
     }
+
+    /// Performs the Protobuf encoding for an enum field.
+    fn encode_enum<T, +Drop<T>, +Into<T, u32>>(
```
rnbguy, line 74: good idea of using `+Into<T, u32>>` for enums. 👍🏼

cairo-libs/packages/protobuf/src/tests/oneof.cairo
```diff
@@ -0,0 +1,127 @@
...
+                ProtobufTag { field_number: 2, wire_type }
+            },
+            Oneof::Enum(e) => {
+                let n: u32 = e.into();
+                n.encode_raw(ref context);
+                let wire_type = ProtoMessage::<u32>::wire_type();
+                ProtobufTag { field_number: 3, wire_type }
+            },
+        }
+    }
```
rnbguy, line 57: I am not sure why this looks like this. IMO it should look like this,

```cairo
        match self {
            Oneof::Int(n) => context.encode_field(1, n),
            Oneof::String(str) => context.encode_field(2, str),
            Oneof::Enum(e) => context.encode_field(3, e),
        }
```

cairo-libs/packages/protobuf/src/tests/oneof.cairo
```diff
@@ -0,0 +1,127 @@
...
+        } else if tag.field_number == 3 {
+            if tag.wire_type != ProtoMessage::<u32>::wire_type() {
+                return Option::None;
+            }
+            let value = context.decode_enum(3)?;
+            Option::Some(Oneof::Enum(value))
+        } else {
+            Option::None
+        }
+    }
```
rnbguy, line 81: we don't need to check for the wire types. It's checked inside `decode_field`


IMO it should look like

```cairo
        match tag.field_number {
          0 => Option::None,
          1 => context.decode_field(1).map(Oneof::Int),
          2 => context.decode_field(2).map(Oneof::String),
          3 => context.decode_field(3).map(Oneof::Enum),
          4 => Option::None,
        }
```

Farhad-Shabani: Good call [17b0b71](https://github.com/informalsystems/ibc-starknet/pull/301/commits/17b0b7188ede8b0e1b44d4294f9ae86abc12267b)

cairo-libs/packages/protobuf/src/tests/oneof.cairo
Farhad-Shabani: The thing is normal protobuf encoding works so that if a field is empty or unavailable ([i.e., it matches the default value](https://github.com/informalsystems/ibc-starknet/blob/17b0b7188ede8b0e1b44d4294f9ae86abc12267b/cairo-libs/packages/protobuf/src/types/message.cairo#L40)), it's ignored and doesn't appear in the result. However, for enums, the default value shouldn’t be ignored.

We might be able to simplify things based on your suggestion, but should the entire Protobuf implementation be refactored (out of this PR's scope) and decoupled from defaults (if possible). Might be better to deal with it as a lower-priority issue later.

Farhad-Shabani: BTW, the alternative to the `Default` is opting for optional fields, as that's how `prost` handles it too.

rnbguy: ah good catch ! 🙌🏼 

but I still think, this ruins the clean code. I see an immediate hack, which is, `context.encode_repeated_field(2, array![str])` which doesn't ignore the default value.

Maybe add one method `encode_field_forced()` to not ignore default values.

rnbguy: It seems, the variant is broke that all the `ProtoMessage` can be used for `encode_field`. I can't use `Enum(e)` in `context.encode_enum(3, e)`. I have to call `context.encode_enum(3, e)`.

cairo-libs/packages/protobuf/src/types/message.cairo
rnbguy: I think, I didn't implement this way, because I wanted to keep the `encode_field` or `encode_repeated_field` as the top method to be called when implementing `ProtoMessage` on local struct. I don't mind having `encode_enum` and `encode_oneof` as long as, it is called inside `encode_field` and can be generalized for `encode_repeated_field` too.

rnbguy (approved):
I am approving this PR as I feel this will unblock you. But my points about `field` vs `oneof` and `enums` stays valid. As you suggested, I will open an issue for this.

cairo-libs/packages/protobuf/src/tests/oneof.cairo
Farhad-Shabani: I get what you're aiming for, but let's address this when it really matters for the project. I dropped my two cents in the issue you opened:  https://github.com/informalsystems/ibc-starknet/issues/323#issuecomment-2685606836

Issue #290, opened 2025-02-18
better strategy at polling events or transaction response


rnbguy:
https://github.com/informalsystems/ibc-starknet/blob/e0e617bfae7d9b155e1eded66c14682e81b46309/relayer/crates/starknet-chain-components/src/impls/queries/block_events/retry.rs#L27

https://github.com/informalsystems/ibc-starknet/blob/e0e617bfae7d9b155e1eded66c14682e81b46309/relayer/crates/starknet-chain-components/src/impls/queries/block_events/wait.rs#L32

These hardcoded values work (mostly) in the local devnet. But in public testnet and mainnet, the block time is 30 secs. So, it doesn't make sense to start `poll_tx_response` right after submitting a transaction.

Also, public RPC endpoints will rate-limit if we use their endpoints too often unnecessarily.

For an example of a better heuristic, we should retry only after `last_committed_block_timestamp + block_time`.

rnbguy:
Agree with your point. Maybe we need to use compile-time environment variables just for us. Otherwise, I can't use the current code with public RPC endpoints.

soareschen:
Let's add a poll interval field on `StarknetChain` to make it configurable then. We can define an auto getter trait like:

```rust
#[cgp_auto_getter]
pub trait HasPollInterval {
    fn polll_interval(&self) -> &Duration;
}
```

Btw, the [first](https://github.com/informalsystems/ibc-starknet/blob/e0e617bfae7d9b155e1eded66c14682e81b46309/relayer/crates/starknet-chain-components/src/impls/queries/block_events/retry.rs#L27) duration probably do not need to be configurable. that is a retry logic which starts at 500ms, and then increase the retry time exponentially in case if the query returns error.

PR #302, opened 2025-02-21
Add `poll_interval` configuration for Starknet chains

ljoss17:
Closes: #290

This PR adds a new configuration for Starknet chains which sets the polling interval

### DIFF
--- a/relayer/Cargo.lock
+++ b/relayer/Cargo.lock
@@ -1825,7 +1825,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 [[package]]
 name = "hermes-any-counterparty"
 version = "0.1.0"
-source = "git+https://github.com/informalsystems/hermes-sdk.git#f39c71bba9e24059a0dd1e1ccfb892d137406d1d"
+source = "git+https://github.com/informalsystems/hermes-sdk.git#aadd03ab7840beff4e9fe0cf42f082171fd73613"
 dependencies = [
  "cgp",
  "hermes-cosmos-chain-components",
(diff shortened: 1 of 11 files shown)


rnbguy:
what happens to `DefaultPollTimeout`?

https://github.com/informalsystems/ibc-starknet/blob/1414bc80cb42ca3116c5e2b7e8c842056ce622b6/relayer/crates/starknet-chain-components/src/impls/tx_response.rs#L57-L68

ljoss17:
> what happens to `DefaultPollTimeout`?
> 
> https://github.com/informalsystems/ibc-starknet/blob/1414bc80cb42ca3116c5e2b7e8c842056ce622b6/relayer/crates/starknet-chain-components/src/impls/tx_response.rs#L57-L68

This seems to be used when querying the Tx response:
* https://github.com/informalsystems/hermes-sdk/blob/main/crates/relayer/relayer-components/src/transaction/impls/estimate_fees_and_send_tx.rs#L79
* https://github.com/informalsystems/hermes-sdk/blob/main/crates/relayer/relayer-components/src/transaction/impls/poll_tx_response.rs#L78-L79

rnbguy:
I need the polling configuration at both. This polling is used when the relayer submits a packet on Starknet and waits for it to get included.

soareschen:
We can don't use `DefaultPollTimeout`, and instead define a provider that reads from the same `poll_interval` field.

ljoss17:
Since both require updating Hermes SDK, is there a preferred solution?

soareschen:
Probably. I have forgotten that `HasPollTimeout` exists. So we can remove `HasPollTimeout::poll_backoff`, and implement it in similar way as `HasPollInterval`.

ljoss17:
`poll_backoff` has been replaced by `poll_interval`, so it is now configurable.
`poll_timeout` uses the `FixedPollTimeout` of 300 seconds.
Let me know if there is an issue with this solution

rnbguy (approved):
👍🏼

PR #303, opened 2025-02-21
fix: printing `call_data_str` in `starknet transfer-args` cli

rnbguy:
I mistakenly removed the `format!()` in #284

caught by `cargo +nightly clippy -- -W clippy::nursery`

Also, added some other touch ups.

- derive `Eq`
- derive `const fn new` for `StarknetAddress`
- use `StarknetAddress` directly while decoding Felt data.


ljoss17 (approved):
Looks good to me!

PR #313, opened 2025-02-25
Improve test error log

ljoss17:
This PR map the `execution.send()` error to be display as Debug in tests

Old error:
```
Error: TransactionExecutionError
```
New error:
```
Error: StarknetError: TransactionExecutionError(TransactionExecutionErrorData { transaction_index: 0, execution_error:
"Transaction execution has failed:\n0: Error in the called contract (contract address:
0x049dfb8ce986e21d354ac93ea65e6a11f639c1934ea253e5ff14ca62eca0f38e, class hash:
0x061dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f, selector:
0x015d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad):\nError at pc=0:4835:\nCairo
traceback (most recent call last):\nUnknown location (pc=0:67)\nUnknown location (pc=0:1835)\nUnknown location
(pc=0:2554)\nUnknown location (pc=0:3436)\nUnknown location (pc=0:4040)\n\n1: Error in the called contract (contract
address: 0x046dcfe2e7f9bbb71dd8f9812b5a9cb76ae7685d3d5af0d0bf11bab2c3f2feab, class hash:
0x0000000000000000000000000000000000000000000000000000000000000000, selector:
0x0219209e083275171774dab1df80982e9df2096516f06319c5c6d71ae0a8480c):\nRequested contract address
0x046dcfe2e7f9bbb71dd8f9812b5a9cb76ae7685d3d5af0d0bf11bab2c3f2feab is not deployed.\n" })
```

### DIFF
--- a/relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
+++ b/relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
@@ -60,6 +60,7 @@ use hermes_starknet_chain_components::types::register::{MsgRegisterApp, MsgRegis
 use hermes_starknet_chain_context::contexts::chain::StarknetChain;
 use hermes_starknet_chain_context::contexts::encoding::cairo::StarknetCairoEncoding;
 use hermes_starknet_chain_context::contexts::encoding::event::StarknetEventEncoding;
+use hermes_starknet_chain_context::impls::error::SignError;
 use hermes_starknet_relayer::contexts::cosmos_to_starknet_relay::CosmosToStarknetRelay;
 use hermes_starknet_relayer::contexts::starknet_to_cosmos_relay::StarknetToCosmosRelay;
 use hermes_test_components::bootstrap::traits::chain::CanBootstrapChain;
@@ -69,7 +70,7 @@ use ibc::core::connection::types::version::Version as IbcConnectionVersion;
 use ibc::core::host::types::identifiers::{PortId as IbcPortId, Sequence};
 use poseidon::Poseidon3Hasher;
 use sha2::{Digest, Sha256};
-use starknet::accounts::{Account, Call, ExecutionEncoding, SingleOwnerAccount};
+use starknet::accounts::{Account, AccountError, Call, ExecutionEncoding, SingleOwnerAccount};
 use starknet::core::types::U256;
 use starknet::macros::{selector, short_string};
 use starknet::providers::Provider;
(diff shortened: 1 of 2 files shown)


rnbguy (approved):
Thanks so much for debugging this ! 👍🏼

rnbguy:
@Farhad-Shabani do you have any idea what may have caused this. It seems, the contract is the erc20 token for the corresponding cosmos contract. But it's deployed and can be queried for the balance (example on CI: https://github.com/informalsystems/ibc-starknet/actions/runs/13522930603/job/37794980266#step:8:630)

Farhad-Shabani:
> @Farhad-Shabani do you have any idea what may have caused this.

I could reproduce the error, which bubbles up some times. I see a difference in one character between the printed token address and the one being called during execution (the example CI link). 

```bash
0x46dcfe2e7f9bbb71dd8f9812b5a9cb76ae7685d3d5af0d0bf11bab2c3f2feab  # ics20 printed address
0x046dcfe2e7f9bbb71dd8f9812b5a9cb76ae7685d3d5af0d0bf11bab2c3f2feab # the one is called
```

It seems like, for some address formats [during this step](https://github.com/informalsystems/ibc-starknet/blob/farhad/impl-proto-oneof/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L517-L535), a leading zero is somehow getting added. Not sure if it's due to the encoding, Poseidon hashing or something wrong with the `execute_v3` API.

Farhad-Shabani:
> The other possibility may be that the `ibc_token_address` function on the Cairo contract is returning an incorrect address?

I don't believe so, since the [stored token address by ICS20](https://github.com/informalsystems/ibc-starknet/blob/17b0b7188ede8b0e1b44d4294f9ae86abc12267b/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L685) is the exact one thrown out by the [`deploy_syscall`](https://github.com/informalsystems/ibc-starknet/blob/17b0b7188ede8b0e1b44d4294f9ae86abc12267b/cairo-contracts/packages/apps/src/transfer/erc20_call.cairo#L53), without any sort of operations on it.

soareschen:
Actually, we may need to increase the timeout [here](https://github.com/informalsystems/ibc-starknet/blob/main/relayer/crates/starknet-integration-tests/src/tests/ics20.rs#L466) to decrease the flakiness. To properly eliminate the sleep, I have filed #320 so that we can poll the address without access to the background relayer.

I am not sure if the failure is caused by Starknet-devnet returning result from a pending state for queries, which does not match the actual state when the transaction is processed.

rnbguy:
FYI, I don't think, it's either of the `Display` impl or incorrect contract logic. I just tried twice locally and both passed with an ics20 token address that is 63 string length in hex.

As Soares mentioned,

> I am not sure if the failure is caused by Starknet-devnet returning result from a pending state for queries, which does not match the actual state when the transaction is processed.

This is probably the reason. Since the packet is auto-relayed in parallel now, there is no way to sync this call. I guess, increasing the sleeping is the only option.

For the demo, I will wait for the block explorer to confirm `Accepted on L2` before proceeding for next txs.

rnbguy:
Or we can create an util method to check if the corresponding packet commitment is received on counterparty before continuing.

PR #317, opened 2025-02-25
fix: cairo contract information in `StarknetChainConfig`

### LINKED ISSUE #308: extend `StarknetChainConfig` with necessary Cairo classes and addresses
Currently, the `StarknetChainConfig` doesn't take the necessary Cairo classes and addresses.

https://github.com/informalsystems/ibc-starknet/blob/7b625a0387bca578a6023c594539629e5cc0fd62/relayer/crates/starknet-chain-components/src/impls/types/config.rs#L15-L20

This prevents auto-relaying from CLI.

We need at least IBC core contract address and Comet client contract address.

https://github.com/informalsystems/ibc-starknet/blob/7b625a0387bca578a6023c594539629e5cc0fd62/relayer/crates/starknet-chain-context/src/contexts/chain.rs#L200-L211

### LINKED ISSUE #315: support for default `StarknetEventEncoding`
We need to pass the class hashes to `StarknetEventEncoding` while running `hermes-starknet` cli.

https://github.com/informalsystems/ibc-starknet/blob/00291960de6ab9ebee352ddbde1a5f65bd397285/relayer/crates/starknet-chain-context/src/contexts/encoding/event.rs#L29-L34

We also need to pass these class hashes in the PR resolving #308.

rnbguy:
closes #315
closes #308

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/types/config.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/types/config.rs
@@ -2,7 +2,9 @@ use core::time::Duration;
 
 use hermes_cosmos_chain_components::impls::types::config::CosmosChainConfig;
 use serde::{Deserialize, Serialize};
+use starknet::core::types::Felt;
 
+use crate::impls::types::address::StarknetAddress;
 use crate::types::wallet::StarknetWallet;
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


soareschen (approved):

relayer/crates/starknet-chain-components/src/impls/types/config.rs
```diff
@@ -17,4 +19,19 @@ pub struct StarknetChainConfig {
     pub relayer_wallet: StarknetWallet,
     #[serde(with = "humantime_serde")]
     pub poll_interval: Duration,
+    pub contract_addresses: StarknetContractAddresses,
+    pub contract_classes: StarknetContractClasses,
```
soareschen, line 23: We can probably refactor the code later to always use contract addresses instead of class hashes. I originally used class hashes, with the assumption that we may switch to permissionless IBC contracts on Starknet. But now that seems increasingly unlikely to happen.

rnbguy: Ah. I wasn't aware. Thanks for pointing this out.

Issue #320, opened 2025-02-25
Provide Cairo methods to pre-allocate ICS20 token address


soareschen:
Currently, the ICS20 address is only allocated after the first IBC transfer of a denom happened. This makes it difficult to write tests or applications to check for the status of the transfer, as we can't query for the balance if we don't know the token address.

The ICS20 Cairo contract should provide an invocation method to pre-allocate a token address, so that a sender can have expectation of what token they would receive when the actual transfer happens.

PR #321, opened 2025-02-25
feat: expose `create_ibc_token` endpoint allowing pre-creation

Farhad-Shabani:
Closes: #320

### DIFF
--- a/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
+++ b/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
@@ -815,7 +815,6 @@ pub mod TokenTransferComponent {
             token_address: ContractAddress,
         ) {
             let denom_key = denom.key();
-            println!("denom_key: {:?}", denom_key); // 🪚
 
             self.write_ibc_token_key_to_address(denom_key, token_address);
 
--- a/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
+++ b/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
@@ -19,8 +19,8 @@ pub mod TokenTransferComponent {
         PrefixedDenom, PrefixedDenomTrait, TracePrefixTrait,
     };
     use starknet_ibc_apps::transfer::{
-        ERC20Contract, ERC20ContractTrait, ISendTransfer, ITransferQuery, ITransferrable,
-        SUCCESS_ACK, TRANSFER_PORT_ID, TransferErrors, VERSION,
+        ERC20Contract, ERC20ContractTrait, ICreateIbcToken, ISendTransfer, ITransferQuery,
+        ITransferrable, SUCCESS_ACK, TRANSFER_PORT_ID, TransferErrors, VERSION,
     };
     use starknet_ibc_core::channel::{
         AckStatus, AckStatusImpl, Acknowledgement, AppVersion, ChannelContract,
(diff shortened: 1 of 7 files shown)


soareschen (approved):

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
```diff
@@ -277,6 +277,29 @@ pub mod TokenTransferComponent {
...
+    #[embeddable_as(CreateIbcToken)]
+    impl CreateTokenImpl<
+        TContractState, +HasComponent<TContractState>, +Drop<TContractState>,
+    > of ICreateIbcToken<ComponentState<TContractState>> {
+        fn create_ibc_token(
+            ref self: ComponentState<TContractState>,
+            chan_id_on_b: ChannelId,
+            base_denom: ByteArray,
+        ) -> ContractAddress {
+            let token = self.create_token(base_denom.clone(), 0);
```
soareschen, line 293: Actually this is incorrect. We need to first check that the token has not been created yet. Otherwise this would override existing minted token and cause the original token to be lost forever.

Also, `record_ibc_token` needs to error if an existing entry already exist, so that it prevent re-entrant attacks or any potential bugs.

rnbguy:
Why this PR is needed ? 🤔 I don't mind this PR for testing -- but this shouldn't be in the final release of Starknet IBC contracts.

Also, I would like that `create_ibc_token` is called only by permissioned wallet.

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
```diff
@@ -277,6 +277,29 @@ pub mod TokenTransferComponent {
...
+    // ICreateToken
+    // -----------------------------------------------------------
+
+    #[embeddable_as(CreateIbcToken)]
+    impl CreateTokenImpl<
+        TContractState, +HasComponent<TContractState>, +Drop<TContractState>,
+    > of ICreateIbcToken<ComponentState<TContractState>> {
+        fn create_ibc_token(
+            ref self: ComponentState<TContractState>,
+            chan_id_on_b: ChannelId,
```
soareschen, line 290: We should also check that the given channel ID is valid, and there is already a corresponding channel that is opened.

soareschen:
> Why this PR is needed ? 🤔 I don't mind this PR for testing -- but this shouldn't be in the final release of Starknet IBC contracts.

This is also essential for the UX for both developer and user facing. If you are transferring a token to Starknet for the first time, how are you supposed to know which address you should check? The user or developer won't be running the relayer. They may not even be familiar with IBC enough to know how to check the status of the relayed packet. With this method, the whole UX is simplified to the fact that I am transferring token X from Cosmos, and I know I just need to check for my balance of token X' on Starknet.

rnbguy:
> The user or developer won't be running the relayer. They may not even be familiar with IBC enough to know how to check the status of the relayed packet.

There is a query method in ics20 contract exactly for this.

https://github.com/informalsystems/ibc-starknet/blob/627a226bb97e5ad765d0d16fc014aed0fc5235ae/cairo-contracts/packages/apps/src/transfer/interfaces/transfer.cairo#L19-L38

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
Farhad-Shabani: There's an incrementing [salt](https://github.com/informalsystems/ibc-starknet/blob/627a226bb97e5ad765d0d16fc014aed0fc5235ae/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L805) for this, ensuring that each generated token gets a fresh, unique address—even if its metadata is identical to an existing one.

rnbguy: Just throwing an idea. Since we are going with `create_ibc_token`, maybe we make this mandatory and disallow token creation inside at ICS20's `recv_packet`?

rnbguy: To allow my above suggestion, we *MUST* pass a `PrefixedDenom` here.

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
```diff
@@ -277,6 +277,29 @@ pub mod TokenTransferComponent {
...
+            base_denom: ByteArray,
+        ) -> ContractAddress {
+            let token = self.create_token(base_denom.clone(), 0);
+            let trace_prefix = TracePrefixTrait::new(TRANSFER_PORT_ID(), chan_id_on_b);
```
rnbguy, line 294: Also, in the current form of `create_ibc_token`, Osmosis can only index tokens native to itself -- `OSMO` and `ION`. The remote IBC tokens, such as, `ATOM` (on Osmosis) can't be created currently.

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
Farhad-Shabani: I'd advocate keeping it within the `recv_packet` since that's exactly how IBC traditionally works. Thus familiar to IBC users and even expected by chains, operators, etc.

Think of this `create_ibc_token` endpoint as more of a QoL feature. Plus, I don't think this unification would add much extra security. Attacks are more likely to happen at the weakest and most obvious points—especially if we see the `create_ibc_token` to be the more vulnerable one.

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
Farhad-Shabani: Good point. Let me see what can be done for this.

PR #322, opened 2025-02-26
fix: pre-create IBC token in test to reduce flakiness

soareschen:
Fixes the issue discussed in #313.

This PR uses the new `create_ibc_token` Cairo contract call introduced in #321 to pre-allocate the token address for tokens transferred from Cosmos to Starknet. This allows us to use the `assert_eventual_amount` method to poll for the transfer status, since we can now know the transferred address beforehand.

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/types/address.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/types/address.rs
@@ -36,7 +36,7 @@ impl<Chain: Async> ProvideAddressType<Chain> for ProvideFeltAddressType {
     Deserialize,
 )]
 #[display("0x{_0:x}")]
-pub struct StarknetAddress(Felt);
+pub struct StarknetAddress(pub Felt);
 
 pub struct EncodeStarknetAddress;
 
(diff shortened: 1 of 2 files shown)


relayer/crates/starknet-chain-components/src/impls/types/address.rs
```diff
@@ -36,7 +36,7 @@ impl<Chain: Async> ProvideAddressType<Chain> for ProvideFeltAddressType {
     Deserialize,
 )]
 #[display("0x{_0:x}")]
-pub struct StarknetAddress(Felt);
```
rnbguy, line 39: why do you need `pub` here? We already impl `Deref`.

relayer/crates/starknet-chain-components/src/impls/types/address.rs
```diff
@@ -36,7 +36,7 @@ impl<Chain: Async> ProvideAddressType<Chain> for ProvideFeltAddressType {
...
+pub struct StarknetAddress(pub Felt);
```
rnbguy, line 39: why do you need `pub` here? We already impl `Deref`.

soareschen: The preferred approach for CGP is to use generics instead of newtypes with private fields to hide internal details. Private fields in crates like ibc-rs have been problematic especially in Rust, due to the excessive needs to provide all kinds of utility methods at the definition side to deal with all possible ways of interaction, e.g. owned/ref/mut/from etc.

Especially in this case, there is no invariant or added benefit of hiding the fact that a Starknet address is a `Felt`. The only reason we have it is to implement a custom `Display` instance, and the only reason we use newtype is that `Display` is not a CGP trait.

rnbguy: It's opinionated, I see 😅 Since, we are using hermes-sdk and cgp, it's ok.

rnbguy (approved):
Thanks 🙏🏼

Farhad-Shabani:
@soareschen, I am working on a complementary changes for this `create_ibc_token`. Let's merge this PR once things are consolidated.

PR #332, opened 2025-02-28
fix: use specific height to query Starknet

### LINKED ISSUE #188: [relayer] query at specific block on starknet
Currently, we use a `Pending` block id to query the latest state.

https://github.com/informalsystems/ibc-starknet/blob/e64a8ecaa708c5c5150b058b6c9bbe1ba9f54d51/relayer/crates/starknet-chain-components/src/impls/contract/call.rs#L21-L38

Ideally, we should have
- `call_contract_with_latest_height` - takes no block id and possibly uses `BlockTag::Pending`
- `call_contract` - takes a block id or `u64`.

---

Note: this is different than #182 which also discusses querying a proof at a specific block.

rnbguy:
closes #188


relayer/crates/starknet-chain-components/src/impls/queries/status_at_height.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::traits::provider::HasStarknetProvider;
+use crate::traits::queries::status_at_height::{
+    ChainStatusAtHeightQuerier, ChainStatusAtHeightQuerierComponent,
+};
+use crate::types::status::StarknetChainStatus;
+
+pub struct QueryStarknetChainStatusAtHeight;
+
+#[cgp_provider(ChainStatusAtHeightQuerierComponent)]
+impl<Chain> ChainStatusAtHeightQuerier<Chain> for QueryStarknetChainStatusAtHeight
```
soareschen, line 17: I think we can use [`CanQueryBlock`](https://github.com/informalsystems/hermes-sdk/blob/main/crates/chain/chain-components/src/traits/queries/block.rs#L11) instead. This is more of a block info. The chain status is used for knowing the current status of the chain.

rnbguy: ac4eaf2

soareschen (approved):

PR #334, opened 2025-02-28
fix: exclude starknet signing key from relayer config file 

### LINKED ISSUE #318: exclude Starknet wallet's signing key in relayer config
Currently, the `StarknetWallet` takes `signing_key` directly.

https://github.com/informalsystems/ibc-starknet/blob/f721ccf98a925b81e897d11d43666df4f033da58/relayer/crates/starknet-chain-components/src/types/wallet.rs#L7-L11

This makes the relayer configuration file include the private secret. Ideally, this should a file path. While creating the `LocalWallet` or other wallets, the secret should be read from it.

This will allow us to include our testnet configuration file in the repo.

rnbguy:
closes #318

included an example relayer config file. after filling out the empty fields, we can relay packets via

```bash
RUST_LOG=trace hermes-starknet -c relayer.toml start cosmos-with-starknet \
  --chain-id-a 393402133025997798000961 --client-id-a 07-tendermint-3 \
  --chain-id-b osmo-test-5 --client-id-b 08-wasm-4459
```

### DIFF
--- a/relayer/crates/starknet-chain-components/src/types/wallet.rs
+++ b/relayer/crates/starknet-chain-components/src/types/wallet.rs
@@ -4,9 +4,10 @@ use starknet::core::types::Felt;
 use crate::impls::types::address::StarknetAddress;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
 pub enum SigningKey {
-    File(String),
-    Felt(Felt),
+    File { file: String },
+    Felt { felt: Felt },
 }
 
 impl TryFrom<SigningKey> for Felt {
@@ -14,19 +15,19 @@ impl TryFrom<SigningKey> for Felt {
 
     fn try_from(value: SigningKey) -> Result<Self, Self::Error> {
         match value {
-            SigningKey::File(path) => Ok(std::fs::read_to_string(path)
+            SigningKey::File { file } => Ok(std::fs::read_to_string(file)
                 .map_err(|e| format!("{e}"))?
                 .trim()
                 .parse()
                 .map_err(|e| format!("{e}"))?),
-            SigningKey::Felt(felt) => Ok(felt),
+            SigningKey::Felt { felt } => Ok(felt),
         }
     }
 }
 
 impl From<Felt> for SigningKey {
     fn from(felt: Felt) -> Self {
-        Self::Felt(felt)
+        Self::Felt { felt }
     }
 }
 
--- a/relayer/crates/starknet-chain-components/src/types/wallet.rs
+++ b/relayer/crates/starknet-chain-components/src/types/wallet.rs
@@ -4,8 +4,35 @@ use starknet::core::types::Felt;
 use crate::impls::types::address::StarknetAddress;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


relayer/crates/starknet-chain-components/src/types/wallet.rs
```diff
@@ -3,9 +3,37 @@ use starknet::core::types::Felt;
...
+    fn from(felt: Felt) -> Self {
+        Self::Felt { felt }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct StarknetWallet {
     pub account_address: StarknetAddress,
-    pub signing_key: Felt,
+    pub signing_key: SigningKey,
```
soareschen, line 37: I think the signing key here should always be `Felt`. We should update the config to load the key from a path, and load the key before initializing the chain or wallet.

relayer/crates/starknet-chain-components/src/types/wallet.rs
```diff
@@ -3,9 +3,37 @@ use starknet::core::types::Felt;
...
+    File { file: String },
+    Felt { felt: Felt },
+}
+
+impl TryFrom<SigningKey> for Felt {
+    type Error = String;
+
+    fn try_from(value: SigningKey) -> Result<Self, Self::Error> {
+        match value {
+            SigningKey::File { file } => Ok(std::fs::read_to_string(file)
```
soareschen, line 18: This is an unexpected side effect, some more with blocking I/O, which would happen every time the field is used, which is not good.

relayer/crates/starknet-chain-components/src/types/wallet.rs
rnbguy: Can you please take over this PR? Maybe I have to do the same way you do it for `CosmosChainConfig`

https://github.com/informalsystems/hermes-sdk/blob/27168d1ee8c004d87b39d769e80c686d79ed53c5/crates/cosmos/cosmos-chain-components/src/impls/types/config.rs#L21-L22

relayer/crates/starknet-chain-components/src/types/wallet.rs
rnbguy: Can you please take over this PR? Maybe I have to do the same way you do it for `CosmosChainConfig`

https://github.com/informalsystems/hermes-sdk/blob/27168d1ee8c004d87b39d769e80c686d79ed53c5/crates/cosmos/cosmos-chain-components/src/impls/types/config.rs#L21-L22

soareschen: No problem, get some rest. I think this is lower priority for now.

relayer/crates/starknet-chain-components/src/types/wallet.rs
rnbguy: 5884df0

soareschen:
@rnbguy I have pushed a commit to properly save and load the wallet files using async I/O. Please test it and see if it works.

rnbguy:
thanks @soareschen ! the changes work on testnet 🙏🏼 we can merge.

soareschen (approved):

Issue #340, opened 2025-03-06
Challenges in obtaining the full denom using an IBC token's address


Farhad-Shabani:
Requested by @ljoss17 , @soareschen 

## Context
From viewpoint of frontend apps on Starknet (and relayers too) there is a challenge where they can see/obtain name of an IBC created token and it’s contract address, but not the full token’s denomination like `transfer/channel-0/uatom`. Thus, It could be tricky to obtain the correct denomination in form of `PrefixedDenom`, and so confidently craft a `MsgTransfer` to send the token back to the source chain. 
It would be nice to have a query endpoint to easier obtain this denom by giving the contract address.

## Remark
The implementation of this endpoint could not be straightforward, the same as existing `ibc_token_address`, just reversed.
A challenge here is that `PrefixedDenom` contains an array of `TracePrefix`, and if we want to return the denom as `PrefixedDenom`, meaning to store it like `Map<ContractAddress, PrefixedDenom>` in storage, we’d need a custom `Store` trait implementation (since it involves arrays). Besides, the issue with arrays is that it can only store up to 255 (`u8::MAX`) elements.

PR #341, opened 2025-03-06
feat: expose `ibc_token_denom` query endpoint

Farhad-Shabani:
Closes: #340

### DIFF
--- a/cairo-contracts/packages/apps/src/tests/transfer.cairo
+++ b/cairo-contracts/packages/apps/src/tests/transfer.cairo
@@ -1,4 +1,4 @@
-use TokenTransferComponent::TransferValidationTrait;
+use TokenTransferComponent::{TransferReaderTrait, TransferValidationTrait, TransferWriterTrait};
 use openzeppelin_testing::events::EventSpyExt;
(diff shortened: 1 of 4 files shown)


ljoss17 (approved):
Looks good to me! Thanks

soareschen (approved):

rnbguy:
Hmm. I am curious -- since we are adding this, does this make sense to store a `ibc_token_denom_to_address: Map<ByteArray, ContractAddress>`?

https://github.com/informalsystems/ibc-starknet/blob/fe59110976f67221c1466e704266338e0b9d0874/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L37-L39

And use it for token query endpoint?

https://github.com/informalsystems/ibc-starknet/blob/fe59110976f67221c1466e704266338e0b9d0874/cairo-contracts/packages/apps/src/transfer/components/transfer.cairo#L327-L335

That way users don't really have to perform a poseidon hash while querying the token address.

soareschen:
I think the issue is that the storage key cannot be something other than `Felt`. But it should be possible to perform the hasing inside the Cairo contract before performing the lookup.

In any case, this can be handled through a different PR.

Issue #337, opened 2025-03-01
`query_packet_acknowledgement` should parse the Ack from block events


rnbguy:
https://github.com/informalsystems/ibc-starknet/blob/83d22c471d003091a1cc5bd4e08b0ba973c33fe7/relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs#L89-L95

These `ack_bytes` are actually acknowledgement commitment -- that is hashed acknowledgement. We incorrectly returned it here

https://github.com/informalsystems/ibc-starknet/blob/83d22c471d003091a1cc5bd4e08b0ba973c33fe7/relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs#L120

We should use `query_block_events` and `StarknetEventEncoding` to parse `PacketRelayEvents::WriteAcknowledgement`. 

This is being used in [`hermes-sdk`](https://github.com/informalsystems/hermes-sdk/blob/ea6d57c405c93d9371f570edf6e5422e10297749/crates/relayer/relayer-components/src/relay/impls/packet_relayers/receive/skip_received_packet.rs#L60-L61).

Ref: https://github.com/informalsystems/ibc-starknet/pull/328#issuecomment-2691163987

PS: `hermes-sdk` also has the [same bug for Cosmos chains](https://github.com/informalsystems/hermes-sdk/blob/ea6d57c405c93d9371f570edf6e5422e10297749/crates/cosmos/cosmos-chain-components/src/impls/queries/packet_acknowledgement.rs#L28-L44).

rnbguy:
`hermes-sdk` can already parse the `WriteAck` event and create `packet_ack`.

https://github.com/informalsystems/hermes-sdk/blob/0ff9d64245fd8e5da0e6e6bcac83888531372c8c/crates/relayer/relayer-components/src/relay/impls/event_relayers/packet_event.rs#L107

This is fixed at informalsystems/hermes-sdk#561 by not using this method while creating `packet_ack`.

ljoss17:
Closed by https://github.com/informalsystems/ibc-starknet/pull/342

PR #342, opened 2025-03-06
Fix ack clearing from Cosmos to Starknet

ljoss17:
Closes: #345 #337

### Cairo

This PR has 2 changes on Cairo side:

* Use nanoseconds instead of seconds for timeout timestamps
* Replace the `packet_data` field with `packet` in the `WriteAcknowledgementEvent`

### Relayer

This PR has 2 changes on the relayer side:

* Improve packet clearing test to assert recv and ack are cleared both ways
* Fix `build_packet_from_write_ack_event`

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs
@@ -12,7 +12,7 @@ use hermes_chain_components::traits::types::height::HasHeightType;
 use hermes_chain_components::traits::types::ibc::{
     HasChannelIdType, HasPortIdType, HasSequenceType,
 };
-use hermes_chain_components::traits::types::packets::ack::HasAcknowledgementType;
+use hermes_chain_components::traits::types::packets::ack::HasAckCommitmentHashType;
 use hermes_chain_components::traits::types::proof::HasCommitmentProofType;
 use hermes_cosmos_chain_components::types::key_types::secp256k1::Secp256k1KeyPair;
 use hermes_encoding_components::traits::decode::CanDecode;
@@ -35,9 +35,6 @@ use crate::types::membership_proof_signer::MembershipVerifierContainer;
 use crate::types::messages::ibc::channel::PortId as CairoPortId;
 use crate::types::messages::ibc::packet::Sequence;
 use crate::types::status::StarknetChainStatus;
-
-const SUCCESS_ACK: &[u8] = br#"{"result":"AQ=="}"#;
-
 pub struct QueryStarknetAckCommitment;
 
 #[cgp_provider(PacketAcknowledgementQuerierComponent)]
@@ -49,7 +46,7 @@ where
         + HasIbcCommitmentPrefix<CommitmentPrefix = Vec<u8>>
         + HasChannelIdType<Counterparty, ChannelId = ChannelId>
         + HasPortIdType<Counterparty, PortId = IbcPortId>
-        + HasAcknowledgementType<Counterparty, Acknowledgement = Vec<u8>>
+        + HasAckCommitmentHashType<AckCommitmentHash = Vec<u8>>
         + HasCommitmentProofType<CommitmentProof = StarknetCommitmentProof>
         + HasBlobType<Blob = Vec<Felt>>
         + HasSelectorType<Selector = Felt>
@@ -65,7 +62,7 @@ where
         + CanDecode<ViaCairo, Product![[u32; 8]]>
         + HasEncodedType<Encoded = Vec<Felt>>,
 {
-    async fn query_packet_acknowledgement(
+    async fn query_packet_acknowledgement_with_proof(
         chain: &Chain,
         channel_id: &ChannelId,
         port_id: &IbcPortId,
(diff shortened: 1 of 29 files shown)


relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -239,11 +240,29 @@ fn test_packet_clearing() -> Result<(), Error> {
             contract_address
         };
 
+        let ics20_contract_address = {
```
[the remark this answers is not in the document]
ljoss17: Oh right, thanks for catching that!

soareschen (approved):

relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs
```diff
@@ -120,14 +117,9 @@ where
...
-
-        // FIXME: we can't send `ack_bytes` as it is stored after hashing.
-        // We should query block event to get the original ack_bytes.
-        //
-        // For now, we just return the SUCCESS_ACK as ack_bytes.
+        // `ack_bytes` is stored after hashing.
+        // query block event is required to get the original ack_bytes.
 
-        Ok((SUCCESS_ACK.to_vec(), dummy_proof))
+        Ok((ack_bytes, dummy_proof))
```
rnbguy, line 123: just wanted to confirm that this method now expects acknowledgment commitment bytes (hashed) at hermes-sdk?

relayer/crates/starknet-chain-components/src/impls/types/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct StarknetChainConfig {
 pub struct StarknetContractAddresses {
     pub ibc_core: Option<StarknetAddress>,
     pub ibc_client: Option<StarknetAddress>,
+    pub ibc_ics20: Option<StarknetAddress>,
```
rnbguy, line 30: so, I should add this field in the `relayer.toml.example` file?

relayer/crates/starknet-chain-components/src/types/cosmos/timestamp.rs
```diff
@@ -19,7 +19,7 @@ where
         value: &Timestamp,
         buffer: &mut Encoding::EncodeBuffer,
     ) -> Result<(), Encoding::Error> {
-        let unix_secs = value.nanoseconds() / 1_000_000_000;
+        let unix_secs = value.nanoseconds();
```
rnbguy, line 22: can you please rename the `unix_secs` variable to `unix_nanos`?

relayer/crates/starknet-chain-components/src/types/cosmos/timestamp.rs
```diff
@@ -35,6 +35,6 @@ where
         buffer: &mut Encoding::DecodeBuffer<'a>,
     ) -> Result<Timestamp, Encoding::Error> {
         let product![unix_secs] = encoding.decode_mut(buffer)?;
-        Ok(Timestamp::from_nanoseconds(unix_secs * 1_000_000_000))
+        Ok(Timestamp::from_nanoseconds(unix_secs))
```
rnbguy, line 38: here as well.

relayer/crates/starknet-chain-components/src/types/messages/ibc/denom.rs
```diff
@@ -45,7 +47,39 @@ impl Display for PrefixedDenom {
...
+                let channel_id = chunk[1].to_string();
+                TracePrefix {
+                    port_id,
+                    channel_id,
+                }
+            })
+            .collect();
+        Ok(Self { trace_path, base })
+    }
+}
```
rnbguy, line 80: this parsing is nuanced.

informalsystems/ibc-rs#1177

Is it possible to use the `ibc-rs` type to parse this? I am aware that, ibc-rs types don't have pub fields -- which can be a blocker. if that's the case, can you add a `FIXME` comment to this ibc-rs issue?

relayer/crates/starknet-integration-tests/src/contexts/bootstrap.rs
```diff
@@ -185,11 +185,13 @@ impl ChainDriverBuilder<StarknetBootstrap> for StarknetBootstrapComponents {
                 account: Arc::new(account),
                 ibc_client_contract_address: None,
                 ibc_core_contract_address: None,
+                ibc_ics20_contract_address: None,
                 event_encoding: Default::default(),
                 proof_signer,
                 poll_interval: core::time::Duration::from_millis(200),
                 block_time: core::time::Duration::from_secs(1),
                 nonce_mutex: Arc::new(Mutex::new(())),
+                signer: relayer_wallet.clone(),
```
rnbguy, line 194: I am curious, how `signer` is different than `account`?

rnbguy:
You are using nanoseconds for Cairo packet timeout. But you forgot to adjust when comparing it with Cairo block time which is in seconds.

There are multiple places this is used. Particularly the `Timestamp` struct is being inside Cairo code -- which is now using seconds and nanoseconds -- which is wrong.

relayer/crates/starknet-chain-components/src/impls/queries/ack_commitment.rs
ljoss17: Currently the ack bytes are not used, only the proof https://github.com/informalsystems/hermes-sdk/blob/main/crates/chain/chain-components/src/impls/payload_builders/packet.rs#L101

relayer/crates/starknet-chain-components/src/impls/types/config.rs
ljoss17: If I'm not mistaken, this is currently only used by `build_ibc_token_transfer_message` to create a transfer. So it's shouldn't be an issue if it is not configured

relayer/crates/starknet-chain-components/src/types/cosmos/timestamp.rs
ljoss17: Opened https://github.com/informalsystems/ibc-starknet/pull/352 to fix this

PR #343, opened 2025-03-06
Test and support bootstrapping of Starknet and Osmosis chain CLI

soareschen:
Merge after: #342, https://github.com/informalsystems/hermes-sdk/pull/568

## Summary

This PR properly tests the bootstrap CLI, and and allows bootstrapping of both Starknet and Osmosis chain for local testing.

I have also included the deployment of the IBC contracts inside `StarknetBootstrap`. This way, we can also remove the copy-pasted code for deploying IBC contracts manually inside tests.

To allow updating of the contract addresses in `StarknetChain`, I have changed the fields to use `OnceLock`, so that the fields can be lazily initialized after the chain context deploys the IBC contracts.

## Testing

You can test the CLI locally by running the provided scripts inside the `relayer` directory:

```bash
./scripts/bootstrap-chain-a.sh
./scripts/bootstrap-chain-b.sh
```

Note that support for the CLIs other than bootstrap are still incomplete. I will implement them in separate PRs.


relayer/crates/starknet-test-components/src/impls/bootstrap/deploy_contracts.rs
```diff
@@ -0,0 +1,363 @@
...
+            logger
+                .log("registered ICS20 contract with ibc core", &LevelInfo)
+                .await;
+        }
+
+        chain
+            .ibc_core_contract_address()
+            .set(ibc_core_address)
+            .map_err(|_| {
+                Bootstrap::raise_error("failed to set ibc_core_contract_address on chain")
```
ljoss17, line 282: Shouldn't the original error be included in the message to avoid losing information?

soareschen: The method only return the original value in `Result`. We are using `OnceLock`, so the value can only be lazily initialized once and then never changed again.

We only perform lazy setting during test, as the chain context needs to be constructed first before deploying the contracts. After that, it won't show up in normal uses.

PR #348, opened 2025-03-10
docs: update documentation resources

rnbguy:


relayer/crates/starknet-cli/examples/bootstrap_ibc.rs
```diff
@@ -297,6 +297,16 @@ fn main() -> Result<(), Error> {
...
+        info!(
+            "Osmosis: client {}, connection {}, channel {}",
+            cosmos_client_id, cosmos_connection_id, cosmos_channel_id
+        );
+
+        info!(
+            "Starknet: client {}, connection {}, channel {}",
+            starknet_client_id, starknet_connection_id, starknet_channel_id
+        );
+
```
rnbguy, line 309: @soareschen any idea why the `info!` logs in this file doesn't actually print anything? maybe the logger is not properly initialized, as this is not really a `#[cfg(test)]`.

soareschen: Probably because by default we filter out all targets don't start with "hermes". You can add `target: "hermes",` inside `info!` to make it pass the filter.

The filter is needed as otherwise going to `DEBUG` or lower will show a lot of logs generated from third party libraries.

soareschen (approved):

scripts/starknet_ibc_core.ts
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env -S deno run
+
+import { Contract, RpcProvider } from "npm:starknet";
+import { QueryClient, setupIbcExtension } from "npm:@cosmjs/stargate";
+import { Tendermint34Client } from "npm:@cosmjs/tendermint-rpc";
+
+const OSMOSIS_RPC_ENDPOINT = "https://rpc.testnet.osmosis.zone";
```
Farhad-Shabani, line 7: Should these come from env?

scripts/starknet_ibc_denom.ts
```diff
@@ -0,0 +1,95 @@
...
+    byteArray,
+    CairoCustomEnum,
+    CallData,
+    Contract,
+    hash,
+    num,
+    RpcProvider,
+} from "npm:starknet";
+
+const STARKNET_RPC_ENDPOINT =
```
Farhad-Shabani, line 13: Same here. Should come from env?

scripts/starknet_ibc_core.ts
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env -S deno run
```
Farhad-Shabani, line 1: Just a side note: My formatter (`prettierd`) made a bunch of changes to these `.ts` files. 
Not sure if you’ve run a formatter on them, but since it’s not part of CI, it should be fine. Though, still, good to keep things formatted.

scripts/starknet_ibc_core.ts
```diff
@@ -0,0 +1,182 @@
...
+            clientLatestHeight,
+        );
+
+    return {
+        portId: PORT_ID,
+        channelId: CHANNEL_ID,
+        connectionId: connectionId,
+        clientId: clientId,
+        clientSequence: clientSequence,
+
```
Farhad-Shabani, line 80:
```suggestion
```

scripts/starknet_ibc_core.ts
```diff
@@ -0,0 +1,182 @@
...
+    const clientState = await queryClient.ibc.client.state(
+        clientId!,
+    );
+
+    return {
+        portId: portId,
+        channelId: channelId,
+        connectionId: connectionId,
+        clientId: clientId,
+
```
Farhad-Shabani, line 123:
```suggestion
```

scripts/starknet_ibc_core.ts
Farhad-Shabani: Perhaps could run `deno fmt`

scripts/starknet_ibc_core.ts
rnbguy: b3914332

scripts/starknet_ibc_core.ts
rnbguy: b3914332

relayer/crates/starknet-cli/examples/bootstrap_ibc.rs
rnbguy: 64186eb

scripts/starknet_ibc_core.ts
rnbguy: I wanted to leave this as it is, because these two scripts are solely for educational purpose or manual debugging of testnet. They are not meant for automated steps.

Note, I can also generate the output from `cargo run -p hermes-starknet-tools-cli starknet transfer-args ...` using these deno scripts -- but I wanted to avoid adding typescript code in our automated process.

Farhad-Shabani: My thought was these values can constantly change (probably at least until the mainnet launch), so this part of the code could easily become outdated / confusing. Otherwise, we’d need to update these values every time we deploy a new set of contracts. Either way, I’ll leave that up to you. Maybe a short comment helps.

Thanks for adding these!

Farhad-Shabani (approved):

scripts/starknet_ibc_core.ts
rnbguy: on second thoughts -- it's a good idea. let's move these variables in `.env` file.

rnbguy: 6e6e702

scripts/starknet_ibc_denom.ts
rnbguy: 6e6e702

Issue #349, opened 2025-03-11
Inappropriate `cometbft::PublicKey` structure for correct Protobuf (de)encoding


Farhad-Shabani:
## Context
As part of issue #211, the CometBFT [`PublicKey`](https://github.com/informalsystems/ibc-starknet/blob/d253a56fd69722dfe5abbc78b04cd39bf4824035/cairo-libs/packages/cometbft/src/types.cairo#L313) struct and its associated fields need to be refactored into an appropriate `Oneof` field representation so that we can correctly perform Protobuf encoding and decoding.
This change is essential to get the necessary artifacts ready before implementing the header verification logic.

PR #351, opened 2025-03-11
refactor: re-define `cometbft::PublicKey` with `Oneof` field

Farhad-Shabani:
Closes: #349

### DIFF
--- a/cairo-libs/packages/cometbft/src/types.cairo
+++ b/cairo-libs/packages/cometbft/src/types.cairo
@@ -341,12 +341,12 @@ impl SumAsProtoOneof of ProtoOneof<Sum> {
         match self {
             Sum::Ed25519(k) => {
                 k.encode_raw(ref context);
-                let wire_type = ProtoMessage::<ByteArray>::wire_type();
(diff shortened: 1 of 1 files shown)


cairo-libs/packages/cometbft/src/types.cairo
```diff
@@ -309,33 +309,62 @@ impl SignedHeaderAsProtoName of ProtoName<SignedHeader> {
...
     }
 }
 
-impl PublicKeyAsProtoName of ProtoName<PublicKey> {
-    fn type_url() -> ByteArray {
-        "PublicKey"
+#[derive(Drop, Default, Debug, Clone, PartialEq, Serde)]
+pub enum Sum {
+    #[default]
+    Ed25519: Array<u8>,
```
Farhad-Shabani, line 335: Moving towards `Array<u8>,` since in a later step, we’ll need to work with bytes for the signature verification.

rnbguy (approved):
LGTM 🙏🏼 Thanks for taking care of this !

PR #353, opened 2025-03-11
ci: optimize integration test runtime

rnbguy:
fix

- correct rust cache paths
- sequential test runner only for macos-arm64
  - getting failure will be faster on linux-amd64 and linux-arm64
- release profile during integration tests for faster runtime 

current bottleneck

- integration tests themselves
- `snforge_scarb_plugin` always gets recompiled by cargo (pulled as `snforge_std` dep)
  - opened software-mansion/setup-scarb#37
- nix fetch for necessary tools -- it recursively tries to pull all their dependencies.
  - using a rust toolchain action and something like `robinraju/release-downloader@v1` would be faster.
  - although we don't have a release binary for simd-ibc-go-v8-wasm - for that we rely on nix

### DIFF
--- a/.github/workflows/hermes.yaml
+++ b/.github/workflows/hermes.yaml
@@ -56,40 +56,9 @@ jobs:
             relayer
             light-client
 
-      - name: Install osmosisd v28.0.1
-        uses: robinraju/release-downloader@v1
-        with:
-          repository: osmosis-labs/osmosis
-          fileName: ${{ matrix.value.osmosis }}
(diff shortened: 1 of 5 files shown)


Farhad-Shabani (approved):
Nice! Thanks, Rano, for the improvements.

As for `macos`, maybe it's enough to run it only on the `main` branch. That way, PRs stay fast, especially when we're dealing with multiple PRs and want to merge them in a specific order while resolving conflicts.

Issue #193, opened 2025-01-14
[cli] Implement create connection command


soareschen:
Depends on: https://github.com/informalsystems/hermes-sdk/issues/522

Issue #194, opened 2025-01-14
[cli] Implement create channel command


soareschen:
Depend on: https://github.com/informalsystems/hermes-sdk/issues/523

rnbguy:
closed by #354

PR #354, opened 2025-03-11
Add CLI support for create client/connection/channel, start, and clear

soareschen:
Closes: #193, #194
Merge after: https://github.com/informalsystems/hermes-sdk/pull/575

## Summary

This PR fixes the `hermes-starknet` CLI and supports using the CLI for the E2E setup from bootstrapping the chain to starting the relayer.

## Testing

You can test the CLI by running the example shell scripts as follows:

```bash
cd relayer
./scripts/reset.sh

# run in separate terminal
./scripts/bootstrap-chain-a.sh

 
# run in separate terminal, *after* chain-a is fully bootstrapped
./scripts/bootstrap-chain-b.sh

./scripts/create-client-a.sh
./scripts/create-client-b.sh
./scripts/create-connection.sh
./scripts/create-channel.sh

# run in separate terminal
./scripts/start.sh

# perform packet clearing if needed
./scripts/clear.sh
```

Note: the example shell scripts have hard-coded parameters, and will not work if the setup slightly changes. We don't provide sophisticated shell scripts here, because the same thing can already be done in Rust, which is much more customizable and robust than complex shell scripts.


relayer/crates/starknet-relayer/src/contexts/builder.rs
```diff
@@ -120,6 +122,56 @@ impl ChainBuilder<StarknetBuilder, Index<1>> for StarknetBuildComponents {
...
+        Ok(
+            build.build_cosmos_to_starknet_relay(
+                src_chain,
+                dst_chain,
+                src_client_id,
+                dst_client_id,
+            ),
+        )
+    }
+}
```
rnbguy, line 173: does this fixes the ordered-pair type issue in `start starknet-with-cosmos` and `start cosmos-with-starknet`?

Also, wanted to confirm if these subcommands are essentially equivalent with each other.

rnbguy (approved):
Thanks for adding this ! 🚀  Will be great help testing the CLIs.

The instructions work on my machine.

Alos, I see you added all the create subcommands -- including `connection`. does this mean, this PR makes #223 redundant?

relayer/crates/starknet-relayer/src/contexts/builder.rs
soareschen: > Also, wanted to confirm if these subcommands are essentially equivalent with each other.

The relay[1, 0] builder is mainly used in the `create cosmos-client` CLI, which creates a Starknet client on Cosmos. For the create connection and channel commands, we currently simply hardcode to require the first chain to always be Starknet to simplify the implementation and interface.

rnbguy: > So we have to rely on the caller to explicitly disambiguate this.

So the both order works. but Hermes-SDK doesn't have the support for chain type validation.

so users may mix up types and Hermes-SDK will try to relay without proper messages and fail.

did I understand correctly?

soareschen: If you mix up the order, you would just get an error. Because the relayer would for example try to use a Starknet chain ID to construct a Cosmos chain, which it would then fail to find the chain from the config.

PR #355, opened 2025-03-12
Update Hermes SDK

soareschen:


ljoss17:
If we remove the 1 thread limitation for MacOS, we can also run the tests in PR

soareschen:
> If we remove the 1 thread limitation for MacOS, we can also run the tests in PR

Good point. Done in f4cda4dc10faafc0d9247d8d5e1f671c2a5d21c5.

ljoss17 (approved):
Looks good to me!

Issue #181, opened 2025-01-10
impl `CanIbcTransferToken<CosmosChain>: StarknetChain`


rnbguy:
Now that, we managed to get ics20 packet relaying worked in all situations between cosmos and starknet, it makes sense to implement `CanIbcTransferToken<CosmosChain>` for `StarknetChain`.

This will reduce the tech debt and simplify the test in `ics20.rs`.

Farhad-Shabani:
Do we know if this involves any changes to the Hermes SDK, or is it solely related to the relayer implementation in this repo?

rnbguy:
I actually don't know. Of course, @soareschen will have the full context.

soareschen:
We should be able make use of the `SendIbcTransferMessage` implementation from Hermes SDK here. The main dependencies are `CanBuildIbcTokenTransferMessage` and `CanCalculateIbcTransferTimeout`.

PR #359, opened 2025-03-13
Refactoring and simplify `ibc_token_transfer` method

soareschen:
Closes: #181.

Merge after: https://github.com/informalsystems/hermes-sdk/pull/578

- Include construction of ERC20 approval message inside `build_ibc_token_transfer_message`.
- Use `ibc_transfer_token` method to slightly simplify ICS20 tests.
- Use native Starknet denom to construct Prefix denom if call to `ibc_token_denom` fails.


relayer/crates/starknet-chain-components/src/impls/messages/ibc_transfer.rs
```diff
@@ -58,91 +56,126 @@ where
...
+
+                match output {
+                    Ok(output) => {
+                        let prefix_denom_str =
+                            encoding.decode(&output).map_err(Chain::raise_error)?;
+
+                        PrefixedDenom::from_str(&prefix_denom_str).map_err(Chain::raise_error)?
+                    }
+                    Err(_) => {
+                        // FIXME: check and verify error is "ICS20: missing token denom", which means
```
ljoss17, line 118: The error could be logged as warning in the meantime

soareschen: I have modified the Cairo contract to return `Option` in 63d947c5fdc822b1d66256d9118fb7768372c8f6, so that we can pattern match on the response in case the denom is not available.

soareschen: This is now done in d91106734d01255c888a589f4baba443d7868afc.

ljoss17 (approved):
Looks good to me!

PR #360, opened 2025-03-13
build: upgrade scarb, cairo, snforge and etc

### LINKED ISSUE #363: upgrade to starknet 0.13.4 with sierra 1.7.0
We were having issues with declaring Sierra 1.6.0 compiled contract classes on testnet using STRK fees. Which is resolved using sierra 1.7.0 (latest scarb 2.11.3).

1.7.0 sierra comes with Starknet 0.13.4 which is running on testnet, and mainnet is due to upgrade EOM March 2025.

Our e2e tests should start using Sierra 1.7.0 compiled cairo contracts and Starknet 0.13.4 with Sierra 1.7.0 support.

ref: https://www.starknet.io/developers/roadmap

rnbguy:
Closes #363

we can deploy the contracts on testnet. ~but we are blocked on `starknet-devnet` which doesn't support sierra 1.7.0 yet.~ we are using development branches from `starknet-devnet` and `starknet-rs` for sierra `1.7.0` and starknet rpc `0.8.0`.

1. using starknet-devnet dev branch for sierra 1.7.0 support
  a. https://github.com/0xSpaceShard/starknet-devnet/issues/680#issuecomment-2684217493
2. starknet-devnet dev branch uses 0.8.0 RPC endpoint
  a. using `dev/jsonrpc_0_8_0 branch` from https://github.com/xJonathanLEI/starknet-rs
3. `starknet-devnet` fails with [`InsufficientResourcesForValidate`](https://github.com/0xSpaceShard/starknet-devnet/blob/c657668906e0186b8b570520e371ac89f247a40f/crates/starknet-devnet-core/src/error.rs#L101-L102) error.
  a. fixed (https://github.com/0xSpaceShard/starknet-devnet/issues/680#issuecomment-2728932845)
  b. ref. https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257#p-2358763-all-three-bounds-are-present-5


.github/workflows/hermes.yaml
```diff
@@ -47,7 +47,16 @@ jobs:
       - name: Build Cairo contracts
         working-directory: ./cairo-contracts
         run: |
-          scarb build -p starknet_ibc_contracts
+          scarb --profile release build -p starknet_ibc_contracts
```
rnbguy, line 50: using `--profile release` to reduce contract class size.

IBC Core contract class with `dev` and `release` profiles are 2.5M and 2M respectively.

relayer/crates/starknet-integration-tests/src/contexts/chain_driver.rs
```diff
@@ -45,6 +47,7 @@ delegate_components! {
         ErrorRaiserComponent: UseDelegate<HandleStarknetChainError>,
         RuntimeTypeProviderComponent: WithType<HermesRuntime>,
         RuntimeGetterComponent: WithField<symbol!("runtime")>,
+        ChainStartupWaiterComponent: WaitChainReachHeight<1>,
```
rnbguy, line 50: I noticed that now `starknet-devnet` takes some seconds to bootup. So, I added this to wait for `starknet` blocks to produce...

relayer/crates/starknet-integration-tests/src/contexts/starknet_bootstrap.rs
```diff
@@ -172,6 +174,9 @@ impl ChainDriverBuilder<StarknetBootstrap> for BuildStarknetChainDriver {
 
         let rpc_client = Arc::new(JsonRpcClient::new(HttpTransport::new(json_rpc_url)));
 
+        // FIXME: `BuildAndWaitChainDriver`` doesn't help as this call doesn't wait for the chain to start
+        runtime.sleep(core::time::Duration::from_secs(5)).await;
+
```
rnbguy, line 179: .. but I had to add a sleep inside `ChainDriverBuilder` impl as we are querying the RPC endpoint inside of it.

soareschen: That's because `BuildAndWaitChainDriver` waits on the chain driver _after_ it is constructed, but here we need to wait for it to call the underlying rpc_client _before_ it is constructed.

Here it is probably simpler to poll for the `chain_id()` in a loop, since when it succeeds the server also becomes ready.

relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -390,15 +377,9 @@ fn test_packet_clearing() -> Result<(), Error> {
...
-            let tx_hash = execution
-                .send()
-                .await
-                .map_err(<StarknetChain as CanRaiseError<AccountError<SignError>>>::raise_error)?
-                .transaction_hash;
-
-            starknet_chain.poll_tx_response(&tx_hash).await?;
+            let _ = starknet_chain
+                .send_messages_with_signer(wallet_starknet_b, &[StarknetMessage::new(call)])
+                .await?;
```
soareschen, line 382: Let's wait for https://github.com/informalsystems/ibc-starknet/pull/359 to merge, which will remove the whole block of calling approve.

relayer/crates/starknet-test-components/src/impls/bootstrap/start_chain.rs
```diff
@@ -48,6 +48,7 @@ where
             "full",
             "--dump-on",
             "block",
+            "--request-body-size-limit=3000000", // 3M; ibc-core contract class is too large
```
soareschen, line 51: Wouldn't this also cause issues in mainnet later on?

relayer/crates/starknet-chain-components/src/impls/send_message.rs
```diff
@@ -58,7 +58,18 @@ where
 
         let execution = account.execute_v3(calls).nonce(*nonce);
 
+        let fee_estimation = execution.estimate_fee().await.map_err(Chain::raise_error)?;
+
+        // starknet v0.13.4 requires all fee bound present.
+        let l1_gas = core::cmp::max(1, fee_estimation.l1_gas_consumed.try_into().unwrap());
```
soareschen, line 64: Let's handle error here with `.map_err(|_| Chain::raise_error("failed to convert felt to u64"))?`, and add `CanRaiseError<&'static str>` to the constraint.

It is so weird that starknet-rs makes the `TryFrom` instance raise `PrimitiveFromFeltError`, but the type is private and not exported.

relayer/crates/starknet-test-components/src/impls/bootstrap/start_chain.rs
rnbguy: I could deploy ibc-core class with 2.2M size on testnet. 
https://sepolia.starkscan.co/tx/0x031650ea44dce487cba07282a5dfae7ae3f9db229394b7577ccff4a79e90dc20

I think, it's just a node configuration.

relayer/crates/starknet-chain-components/src/impls/send_message.rs
rnbguy: 8ebd61d

relayer/crates/starknet-integration-tests/src/contexts/starknet_bootstrap.rs
rnbguy: d762a09

.github/workflows/hermes.yaml
```diff
@@ -56,21 +65,21 @@ jobs:
...
 
           nix shell \
             ..#osmosis \
             ..#wasm-simapp \
-            ..#starknet-devnet \
             ..#rust \
             ..#cargo-nextest \
             -c \
             cargo nextest run --release \
+            --retries 2 \
```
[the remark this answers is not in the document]
soareschen: The flakiness seems to be caused by either slow start or imprecise timing inside tests, which are worsen on the CI due to slower VMs. We should be able to get rid of the flakiness later, once we refactor the tests and use heights instead of time to compare timings like timeout.

soareschen (approved):

Issue #179, opened 2025-01-10
use `Vec<u8>` instead of `Vec<Felt>` for consistency with Cairo contract


rnbguy:
In cairo code, we use `Array<u8>` for `StateProof` and `Acknowledgment`

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/cairo-contracts/packages/core/src/commitment/types.cairo#L103-L105

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/cairo-contracts/packages/core/src/channel/types.cairo#L320-L322

But because of #178, in the rust relayer code, we use `Vec<Felt>` for `StateProof` and `Acknowledgment`.

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/starknet-chain-components/src/types/messages/ibc/packet.rs#L77-L81

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/starknet-chain-components/src/types/messages/ibc/packet.rs#L143-L146

Issue #178, opened 2025-01-10
`EncodeByteArray` should encode to `String` and not to `Vec<u8>`


rnbguy:
We have

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/cairo-encoding-components/src/components/encode_mut.rs#L23

IMO, this should be

```rs
(ViaCairo, Vec<u8>): EncodeList,
```

to correspond with `Array<u8>` in Cairo.

If I understood correctly, `ByteArray` in Cairo (loosely) represents byte-string representation. This is already done by

https://github.com/informalsystems/ibc-starknet/blob/c13b888f907494910de113845fbbe559d50112bf/relayer/crates/cairo-encoding-components/src/components/encode_mut.rs#L29

PR #362, opened 2025-03-17
refactor: use `Vec<u8>` for `StateProof`

ljoss17:
Closes: #178 
Closes: #179

This PR has the following changes:

### Cairo side

* Changes the `root` field of `StateRoot` from `ByteArray` to `Array<u8>`
* Adds a custom implementation of `Store` for `Array<u8>`

### Relayer side

* Changes encoding for `Vec<u8>` to use `EncodeList` instead of `EncodeByteArray`
* Changes `proof` field of `StateProof` from `Vec<Felt>` to `Vec<u8>`
* Changes `ack` field of `Acknowledgement ` from `Vec<Felt>` to `Vec<u8>`


cairo-contracts/packages/core/src/commitment/types.cairo
```diff
@@ -155,3 +151,56 @@ pub impl StateRootZero of Zero<StateRoot> {
         self.root.len() > 0
     }
 }
+
+pub impl StoreU8Array of Store<Array<u8>> {
```
Farhad-Shabani, line 155: Since this is for the root, and the root is used in hot paths, let’s add a few unit tests:
- Roundtrip (de)serialization
- Store some roots and check if reading the state matches the expected result

cairo-contracts/packages/core/src/commitment/types.cairo
```diff
@@ -155,3 +151,56 @@ pub impl StateRootZero of Zero<StateRoot> {
...
+    fn read_at_offset(
+        address_domain: u32, base: StorageBaseAddress, mut offset: u8,
+    ) -> SyscallResult<Array<u8>> {
+        let mut arr: Array<u8> = array![];
+
+        let len: u8 = Store::<u8>::read_at_offset(address_domain, base, offset)
+            .expect('Storage Span too large');
+        offset += 1;
+
+        let exit = 2 * len + offset;
```
Farhad-Shabani, line 173: I feel like this should be:

```suggestion
        let exit = len + offset;
```

cairo-contracts/packages/core/src/commitment/types.cairo
```diff
@@ -133,25 +133,19 @@ pub impl StateProofZero of Zero<StateProof> {
...
-    fn into(self: ByteArray) -> StateRoot {
-        StateRoot { root: self }
-    }
+    pub root: [u32; 8],
 }
 
 pub impl StateRootZero of Zero<StateRoot> {
     fn zero() -> StateRoot {
-        StateRoot { root: "" }
+        StateRoot { root: [0; 8] }
```
rnbguy, line 141: good catch !

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
```diff
@@ -73,10 +74,23 @@ where
             chain_id: payload.client_state.chain_id,
         };
 
+        // Convert Vec<u8> to a slice of 8 u32 values.
+        let mut root_u32: Vec<u32> = Vec::new();
+        for chunk_u8 in root.chunks_exact(4) {
+            let value_u32 = (chunk_u8[0] as u32)
+                | ((chunk_u8[1] as u32) << 8)
+                | ((chunk_u8[2] as u32) << 16)
+                | ((chunk_u8[3] as u32) << 24);
```
rnbguy, line 83: probably we can use `from_le_bytes` here. but it takes `[u8; 4]`.

cairo-contracts/packages/core/src/commitment/types.cairo
```diff
@@ -133,25 +133,19 @@ pub impl StateProofZero of Zero<StateProof> {
...
 #[derive(Clone, Debug, Drop, PartialEq, Serde, starknet::Store)]
 pub struct StateRoot {
-    pub root: ByteArray // TODO: Determine the correct type (ByteArray or Array<u8>) once implemented membership proof verification.
-}
-
-pub impl ByteArrayIntoRoot of Into<ByteArray, StateRoot> {
-    fn into(self: ByteArray) -> StateRoot {
-        StateRoot { root: self }
-    }
+    pub root: [u32; 8],
```
rnbguy, line 136: add a comment why `[u32; 8]` was preferred over `Array<u8>`.

note that in general, we may have a state root bigger than 256 bytes.

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
rnbguy: ah you fixed it in 7d99039.

ljoss17: Done in 7d99039a36e90d1827be4b9a033de9a3172a6b1c
But if the root is not of 32 elements, there will be an issue when comparing it with the [u32; 8] hash in Cairo, no?

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
```diff
@@ -73,10 +74,21 @@ where
             chain_id: payload.client_state.chain_id,
         };
 
+        // Convert Vec<u8> to a slice of 8 u32 values.
+        let mut root_u32: Vec<u32> = Vec::new();
```
rnbguy, line 78: actually you don't even have to maintain this `root_u32: Vec`.

```rs
if root.len() % 4 != 0 {
  return Chain::raise_error(SOME_ERROR_MESSAGE);
}

let mut root_slice: [u32; 8] = [0; 8];

for (slice_u32, chunk) in root_slice.zip(root.chunks(4)) {
  *slice_u32 = u32::from_le_bytes([chunk[0], chunk[1], chunk[2], chunk[3]])
}
```

cairo-contracts/packages/core/src/commitment/types.cairo
```diff
@@ -133,25 +133,24 @@ pub impl StateProofZero of Zero<StateProof> {
...
-
-pub impl ByteArrayIntoRoot of Into<ByteArray, StateRoot> {
-    fn into(self: ByteArray) -> StateRoot {
-        StateRoot { root: self }
-    }
+    // [u32; 8] is used over Array<u8> as it doesn't require manual
+    // implementation of `starknet::Store` and for Tendermint clients
+    //
+    // In addition, Array<u8> would require a conversion to [u32; 8] when
+    // verifying membership, since Cairo's sha256 hasher output [u32; 8]
```
rnbguy, line 140:
```suggestion
    // verifying membership, since Cairo's sha256 hasher output is [u32; 8]
    // where each u32 represents the hash bytes in little endian format.
```

rnbguy (approved):
🙏🏼

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
rnbguy: make sure `root.len()` is exactly 256 -- otherwise, it will not fill the `root_slice` correctly.

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
```diff
@@ -73,10 +74,21 @@ where
...
+        for chunk in root.chunks_exact(4) {
```
soareschen, line 79: The root should be converted from `Vec<u8>` to `[u8; 32]` first using `TryFrom`, before converting again to `[u32; 8]`. With that, we can also use simpler loops like `for i in 0..8` and access the fixed size array directly. 

Also, we should move this logic to a dedicated function, and include documentation about how the conversion is done. i.e. using little endian to pack 4 `u8` into `u32`.

relayer/crates/starknet-chain-components/src/types/cosmos/consensus_state.rs
```diff
@@ -10,7 +10,7 @@ use hermes_encoding_components::traits::transform::Transformer;
 #[derive(Debug, HasField)]
 pub struct CometConsensusState {
     pub timestamp: u64,
-    pub root: Vec<u8>,
+    pub root: [u32; 8],
```
soareschen, line 13: Shouldn't `CometUpdateHeader` also be updated accordingly?

soareschen (changes requested):

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
```diff
@@ -73,10 +74,21 @@ where
...
+            let chunk_u8: [u8; 4] = chunk.try_into().expect("Chunks must have 4 elements");
+            let value_u32 = u32::from_le_bytes(chunk_u8);
```
Farhad-Shabani, line 81: The calculated root by ICS-23 which we verify this root against uses big-endian, so our implementation here must follow the same convention. Generally, as a rule of thumb, all on-chain bytes are processed in big-endian format.

cairo-contracts/packages/core/src/commitment/types.cairo
ljoss17: Done in 14a2e0b6605272f70210e1c8fe8fafabf2003e92

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
ljoss17: Done in ee085c68f267ab8c7fce9b099ef4da7084a9fd64

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
ljoss17: Done in ee085c68f267ab8c7fce9b099ef4da7084a9fd64

relayer/crates/starknet-chain-components/src/types/cosmos/consensus_state.rs
ljoss17: Done in ee085c68f267ab8c7fce9b099ef4da7084a9fd64

relayer/crates/starknet-chain-components/src/impls/messages/create_client.rs
ljoss17: Done in ee085c68f267ab8c7fce9b099ef4da7084a9fd64

relayer/crates/starknet-chain-components/src/impls/utils/array.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    let mut u32_slice: [u32; 8] = [0; 8];
+
+    for (i, u32_value) in u32_slice.iter_mut().enumerate() {
+        let offset = i * 4;
+        *u32_value = u32::from_be_bytes([
+            u8_slice[offset],
+            u8_slice[offset + 1],
+            u8_slice[offset + 2],
+            u8_slice[offset + 3],
```
rnbguy, line 17: you can do `u32::from_be_bytes(u8_slice[(i*4)..(i + 1)*4])`

relayer/crates/starknet-chain-components/src/impls/utils/array.rs
```diff
@@ -0,0 +1,22 @@
...
+        ]);
+    }
+
+    Ok(u32_slice)
+}
```
rnbguy, line 22: can you add a test here following
https://docs.swmansion.com/scarb/corelib/core-sha256.html

soareschen (approved):

PR #365, opened 2025-03-19
chore: update issue templates

rnbguy:
add github issue templates into the repo.

ref: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

we should also add PR templates sometime in future.


Farhad-Shabani (commented):
I appreciate the comprehensive approach with all these template sections, but IMHO we should prioritize simplicity to make opening issues a breeze for developers while still capturing the information we need. 
Maybe we could merge a few sections? `Reproduction Steps`, `Expected Behavior`, and `Additional Context` could all be combined into the main description, with brief explanations for each included there. The same for the feature request template too. Specifically suggesting that, since we're not dealing with external contributors right now, a streamlined template would reduce friction for our team.

ljoss17 (approved):

PR #361, opened 2025-03-14
feat: implement Ed25519 signature verification

Farhad-Shabani:
Closes: #350

PR #366, opened 2025-03-19
feat: cometbft header verifier library

### LINKED ISSUE #367: implement cometbft header verifier types and methods
now that we have ed25519 signature verification (#350), we need to implement the other validation logic and complete the cometbft light client verifier cairo package.

Farhad-Shabani:
Closes #367

## Description

This PR includes my most recent local changes toward implementing CometBFT header verification. It provides the scaffolding types and functions needed for the verifier, though it's not yet complete. There are still some other underlying functions and types that need to be defined.

You can think of this as a continuation of PR #361, as it aims to integrate the signature verifier into the header verification process. The logic for these parts is included, but should be considered a first pass that definitely needs to be validated with unit tests.

Up to the team whether you want to continue implementation on top of this PR, break it down into smaller changes, or any other approach that better suits your progress.
cc @rnbguy @mpoke 

## Remark
I initially scaffolded everything in a new separate `cometbft_verifier` library, similar to how it exists as a standalone package under `tendermint-rs`. However, after consideration, I thought this might create unnecessary compilation overhead and complexity for our needs. For Starknet, I think the entire verifier implementation can live within the `cometbft` library, which should meet all our requirements.


cairo-libs/packages/cometbft/tests/verifier.cairo
```diff
@@ -0,0 +1,85 @@
+use cometbft::light_client::Header as LcHeader;
+use cometbft::types::{Header as TmHeader, Options, TrustedBlockState, UntrustedBlockState};
+use cometbft::utils::TWO_THIRDS;
+use cometbft::verifier::{header_matches_commit, verify_update_header};
+use protobuf::types::message::ProtoCodecImpl;
+use protobuf::types::wkt::{Duration, Timestamp};
+use protobuf::{base64, hex};
+
+#[test]
+fn test_verify_update_header() {
```
rnbguy, line 10: Managed to get tendermint light client verified !

This test called the top `verify_update_header` 🙂

The header data is captured directly from hermes-sdk update client payload.

rnbguy:
Remaining

- [ ] few more tests cases from tendermint-rs repo
- [ ] code cleaup

cairo-libs/packages/protobuf/src/primitives/utils.cairo
```diff
@@ -35,6 +35,63 @@ pub fn decode_2_complement_32(value: @u32) -> i32 {
     }
 }
 
+pub fn u64_to_little_endian(value: @u64) -> [u8; 8] {
```
rnbguy, line 38: need to add tests for these two methods.

rnbguy: 5b6d294

cairo-libs/packages/cometbft/src/types.cairo
```diff
@@ -472,3 +526,348 @@ impl ValidatorSetAsProtoName of ProtoName<ValidatorSet> {
...
+            return Option::None;
+        }
+    }
+}
+
+#[derive(Drop, Debug, Clone)]
+pub struct UntrustedBlockState {
+    pub signed_header: SignedHeader,
+    pub validators: ValidatorSet,
+    // pub next_validators: Option<ValidatorSet>,
```
rnbguy, line 774: this field is present in `tendermint-rs` but never used in IBC and always set to `None` in ibc-rs.

I removed this, as `Option<T>` implements `Copy`, not `Clone`. And `ValidatoSet` only impls `Clone`.

cairo-libs/packages/cometbft/src/verifier.cairo
```diff
@@ -0,0 +1,308 @@
...
+    verify_validator_sets(@untrusted);
+    validate_against_trusted(@untrusted, @trusted, @options, now);
+    verify_commit_against_trusted(untrusted, trusted, options)
+}
+
+pub fn verify_validator_sets(untrusted: @UntrustedBlockState) {
+    validator_sets_match(untrusted.validators, untrusted.signed_header.header.validators_hash);
+    // validator_sets_match(
+    //     untrusted.next_validators, untrusted.signed_header.header.next_validators_hash,
+    // );
```
ljoss17, line 44: Why don't we check next validators?

cairo-libs/packages/protobuf/src/types/message.cairo
```diff
@@ -96,6 +96,15 @@ pub impl EncodeContextImpl of EncodeContextTrait {
...
         self.buffer.append(@context2.buffer);
     }
+
+    fn encode_with_length<T, +ProtoMessage<T>, +Default<T>, +Drop<T>, +PartialEq<T>>(
+        ref self: EncodeContext, value: @T,
+    ) {
+        let mut context2 = Self::new();
+        value.encode_raw(ref context2);
+        context2.buffer.len().encode_raw(ref self);
+        self.buffer.append(@context2.buffer);
```
ljoss17, line 106: Any reason for naming it `context2`?
```suggestion
        let mut context = Self::new();
        value.encode_raw(ref context);
        context.buffer.len().encode_raw(ref self);
        self.buffer.append(@context.buffer);
```

cairo-libs/packages/cometbft/src/types.cairo
```diff
@@ -401,10 +405,11 @@ impl SumAsProtoOneof of ProtoOneof<Sum> {
...
 #[derive(Default, Debug, Clone, Drop, PartialEq, Serde)]
 pub struct Validator {
-    pub address: ByteArray,
+    pub address: AccountId,
     pub pub_key: PublicKey,
-    pub voting_power: i64,
-    pub proposer_priority: i64,
+    pub voting_power: u64,
+    pub proposer_priority: u64,
+    // pub name: Option<ByteArray>, // not present in protobuf
```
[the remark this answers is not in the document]
rnbguy: Farhad created the first version of these structs from `tendermint-rs` domain types -- which has this field. But in Cairo, I am using the structs for protobuf messages as well -- which doesn't have this field.

I guess, we can remove this.

cairo-libs/packages/cometbft/src/types.cairo
```diff
@@ -472,3 +525,402 @@ impl ValidatorSetAsProtoName of ProtoName<ValidatorSet> {
...
+    validator_address: AccountId,
+    signature: Array<u8> // TODO: whether to define a Signature type?
+}
+
+#[derive(Drop, Debug, Clone, PartialEq, Default)]
+pub struct CanonicalVote {
+    /// Type of vote (prevote or precommit)
+    pub vote_type: VoteType,
+    /// Block height
+    pub height: i64,
```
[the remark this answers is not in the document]
rnbguy: I am not sure, why this was made to `u64`. `i64` and `u64` have different encoding in protobuf. e.g. encoding of `1u64` and `1i64` will be totally different.

maybe we can have protobuf types and then domain types (which may validate and maintain only `u64`) -- but for now, let's maintain a single version of the types.

cairo-libs/packages/cometbft/src/verifier.cairo
rnbguy: next validator field is removed.

check https://github.com/informalsystems/ibc-starknet/pull/366/files#r2022718374

cairo-libs/packages/protobuf/src/types/message.cairo
rnbguy: good catch ! I was copy-pasting 😄

rnbguy: Ah. I just realized that I was using `context2` to signify it is a second context or sub-context -- `self` being the first context or parent-context.

cairo-libs/packages/cometbft/src/types.cairo
```diff
@@ -472,3 +525,402 @@ impl ValidatorSetAsProtoName of ProtoName<ValidatorSet> {
...
+        }
+
+        let mut eq = true;
+        let mut i = 0;
+        let len = lhs_span.len();
+
+        while i < len {
+            if lhs_span.at(i) != rhs_span.at(i) {
+                eq = false;
+                break;
```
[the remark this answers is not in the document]
rnbguy: good catch! in old Cairo version, we couldn't return within a loop. It seems, it is allowed in the latest version.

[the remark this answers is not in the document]
rnbguy: 0f2f64e

cairo-libs/packages/cometbft/src/types.cairo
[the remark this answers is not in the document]
rnbguy: eb7e9d7

ljoss17 (approved):
Looks good to me!

PR #368, opened 2025-03-21
Implement `StarknetSetup` to automate setting up chains and IBC for testing

soareschen:
Closes: #174

Merge after: https://github.com/informalsystems/hermes-sdk/pull/581


relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
```diff
@@ -1,1013 +0,0 @@
...
-            starknet_commitment2.is_none(),
-            "starknet_commitment2 expected to be None"
-        );
-
-        Ok(())
-    })
-}
-
-#[test]
-fn test_relay_timeout_packet() -> Result<(), Error> {
```
ljoss17, line 585: Could we keep this test until it’s migrated to the new format?

relayer/crates/starknet-integration-tests/src/tests/update_clients.rs
```diff
@@ -1,228 +0,0 @@
...
-use hermes_starknet_chain_context::contexts::encoding::cairo::StarknetCairoEncoding;
-use hermes_starknet_relayer::contexts::starknet_to_cosmos_relay::StarknetToCosmosRelay;
-use hermes_test_components::bootstrap::traits::chain::CanBootstrapChain;
-use ibc::core::client::types::Height as CosmosHeight;
-use tracing::info;
-
-use crate::utils::{init_starknet_bootstrap, load_wasm_client};
-
-#[test]
-fn test_relay_update_clients() -> Result<(), Error> {
```
ljoss17, line 30: Could we keep this test until we finish the ICS23 implementation?

relayer/crates/starknet-integration-tests/src/tests/clear_packets.rs
soareschen: Added them back now.

ljoss17 (approved):
Looks good to me!

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
```diff
@@ -289,7 +289,10 @@ pub mod TokenTransferComponent {
             ref self: ComponentState<TContractState>, denom: PrefixedDenom,
         ) -> ContractAddress {
             let mut token = self.get_token(denom.key());
-            assert(token.is_zero(), TransferErrors::TOKEN_ALREADY_EXISTS);
+
+            if !token.is_zero() {
+                return token.address;
```
soareschen, line 294: Here I make token creation to be no-op if a token already exists. This helps resolving the race condition where we try to create a token while a relayer is relaying and triggering the token creation at the same time.

relayer/crates/starknet-chain-components/src/impls/queries/token_address/create.rs
```diff
@@ -0,0 +1,68 @@
...
+            .map_err(Chain::raise_error)?;
+
+        let message = StarknetMessage {
+            call: Call {
+                to: ics20_contract_address.0,
+                selector: selector!("create_ibc_token"),
+                calldata,
+            },
+            counterparty_height: None,
+        };
```
rnbguy, line 58: I see that you now allow returning existing token. Note that, this is a contract invoke -- so we will spend gas on it, even if we don't modify the state. Don't you think it makes sense to query via `ibc_token_address` first ? And create the token if it doesn't exist ?

rnbguy: Ignore I see that, you're doing exactly this later. 👍🏼

soareschen: Yeah, here I make token creation to be no-op if a token already exists. This helps resolving the race condition where we try to create a token while a relayer is relaying and triggering the token creation at the same time.

relayer/crates/starknet-integration-tests/src/tests/test_suite.rs
```diff
@@ -0,0 +1,50 @@
+use cgp::core::field::Index;
+use hermes_cosmos_integration_tests::init::init_test_runtime;
+use hermes_error::types::Error;
+use hermes_ibc_test_suite::tests::clearing::TestPacketClearing;
+use hermes_ibc_test_suite::tests::transfer::TestIbcTransfer;
```
rnbguy, line 5: nice ! ✨

rnbguy (approved):

cairo-contracts/packages/apps/src/transfer/components/transfer.cairo
rnbguy: The race condition makes sense. It's alright for me.

PR #385, opened 2025-04-02
fix: increase ibc transfer timestamp timeout

### LINKED ISSUE #386: `ubuntu-arm` runner fails on e2e tests
Looks like the packet timeout and polling durtion is not enough for `ubuntu-arm` runner.

```
2025-04-02T06:29:35.990247Z ERROR failed to relay packet packet=seq:2, path:channel-0/transfer->channel-0/transfer, toh:no timeout height, tos:Timestamp(2025-04-02 6:29:27.371683553)) src_chain_id=393402133025997798000961 dst_chain_id=chain-1-1265014002 error=Status { code: Unknown, message: "failed to execute message; message index: 1: receive packet verification failed: packet timeout elapsed: current timestamp: 1743575374088392421, timeout timestamp 1743575367371683553: timeout elapsed [cosmos/ibc-go/v8@v8.5.2/modules/core/04-channel/types/timeout.go:35] with gas used: '119626'", metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "288"} }, source: None }
```

Currently, it's set to 90 seconds. We need to bump it higher.

rnbguy:
fixes #386

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/assert/assert_duration.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/assert/assert_duration.rs
@@ -1,0 +1,23 @@
+use core::time::Duration;
+
+use cgp::prelude::*;
+use hermes_test_components::chain::traits::assert::poll_assert::{
+    PollAssertDurationGetter, PollAssertDurationGetterComponent,
+};
+
+pub struct ProvidePollAssertDuration<const INTERVAL: u64, const ATTEMPTS: u32>;
+
+#[cgp_provider(PollAssertDurationGetterComponent)]
+impl<Chain, const INTERVAL: u64, const ATTEMPTS: u32> PollAssertDurationGetter<Chain>
+    for ProvidePollAssertDuration<INTERVAL, ATTEMPTS>
+where
+    Chain: Async,
+{
+    fn poll_assert_interval(_chain: &Chain) -> Duration {
+        Duration::from_secs(INTERVAL)
+    }
+
+    fn poll_assert_attempts(_chain: &Chain) -> u32 {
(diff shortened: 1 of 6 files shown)


rnbguy:
https://github.com/informalsystems/ibc-starknet/blob/688e22edd6f832e06a1dca5b63c58aa767661a3f/relayer/crates/starknet-chain-components/src/components/chain.rs#L586-L587

The default polling is for 90 secs. We need to increase it for slow `ubuntu-arm` runner.

relayer/crates/starknet-chain-components/src/impls/assert/assert_duration.rs
```diff
@@ -0,0 +1,23 @@
...
+    Chain: Async,
+{
+    fn poll_assert_interval(_chain: &Chain) -> Duration {
+        Duration::from_secs(INTERVAL)
+    }
+
+    fn poll_assert_attempts(_chain: &Chain) -> u32 {
+        ATTEMPTS
+    }
+}
```
rnbguy, line 23: probably this should go in `hermes-sdk`.

ljoss17: We can migrate it when required

ljoss17 (approved):

rnbguy:
Ah looks like, the `cosmos-sdk` chain also needs to bump their timeout.

https://github.com/informalsystems/hermes-sdk/blob/8e9d874db6ce2ac32ff68836cc2c84acc260d244/crates/cosmos/cosmos-chain-preset/src/presets/chain.rs#L500-L501

Issue #346, opened 2025-03-10
Use Madara instead of Starknet-Devnet for local E2E tests


soareschen:
## Summary

We currently use starknet-devnet to run the local E2E tests. However, starknet-devnet does not support storage proofs, which is crucial for the next stage of integration.

In this issue, we will attempt to replace starknet-devnet with Madara to run the E2E tests. Note that we won't be using the storage proofs in this task yet. The main goal is to ensure that the existing E2E tests work with Madara, and make any necessary changes that are specific to starknet-devnet.

## Things to Look Out

The following things may need to be changed when integrating the E2E tests with Madara:

- [ ] Implement a Nix package for Madara, or find ways to download pre-build version of Madara on CI, if available.
- [ ] Change the bootstrap component to call the Madara CLI when bootstrapping Starknet.
- [ ] The predefined wallets may differ, so the hardcoded bootstrapped wallets may need to be updated.

rnbguy:
do you know if Madara supports sierra 1.7.0 ? I am trying to move to [starknet 0.13.4](https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257) and 0.8.0 rpc spec in #360.

soareschen:
What version of Cairo does that corresponds to? Seems like Madara is still stuck on Cairo v2.8.4.

rnbguy:
before #360 we were using Cairo 2.9.2 and now we have 2.11.2 on `main`.

PR #388, opened 2025-04-04
Test Madara chain client with basic ERC20 token transfer

soareschen:
Part of: #346

## Summary

This PR adds a bit more testing of the integration of Madara chain with a basic ERC20 contract.

## New `madara-contracts` sub-project

Due to Madara not supporting the latest Cairo version, I have created a minimal `madara-contracts` subproject that can be compiled with earlier versions of Cairo.

Based on the tests in this PR, Madara can at most support Cairo v2.9.4, and any newer version of Cairo contracts will result in Internal Server Error when deploying the contract.

## Incompatible `starknet_traceTransaction` API

Madara has an incompatible version of `starknet_traceTransaction` in its returned response, as reported by https://github.com/xJonathanLEI/starknet-rs/issues/514. In short, it does not wrap the response in the standard Starknet response format, and result in errors when decoding the response.

As a workaround, we have to add explicit JSON RPC implementation that use `reqwest::Client` directly to make the JSON RPC call, so that we can manually parse the non-compliant response returned from Madara.

## Testing

This PR has not yet integrate the new Madara test on the CI. Doing so would require more non-trivial effort, as we would need to find ways to efficiently build Madara on the CI first.

For now, the test can be tested manually with the following steps:

- Build the main branch of Madara at https://github.com/madara-alliance/madara and export the executable in `$PATH`.

- Install Scarb v2.9.4, and run `scarb build` at `madara-contracts`:

```bash
cd madara-contracts
scarb build
```

- Run the Madara bootstrap test with the following command:
```bash
cargo test -p hermes-starknet-madara-tests -- --include-ignored --nocapture test_madara_bootstrap
```


relayer/crates/starknet-madara-tests/src/impls/queries/tx_response.rs
```diff
@@ -49,3 +53,71 @@ where
...
+        let params = TraceTransactionRequestRef {
+            transaction_hash: tx_hash.as_ref(),
+        };
+
+        let request_body = JsonRpcRequest {
+            id: 1,
+            jsonrpc: "2.0",
+            method: JsonRpcMethod::TraceTransaction,
+            params,
+        };
```
rnbguy, line 83: can't we use https://docs.rs/starknet/0.13.0/starknet/providers/trait.Provider.html#tymethod.trace_transaction here?

soareschen: We can't, because the response returned by Madara does not follow the expected format, and result in "data did not match any variant of untagged enum JsonRpcResponse". See https://github.com/xJonathanLEI/starknet-rs/issues/514.

rnbguy: even with `starknet_v13`? 😅 wow

rnbguy (approved):
👌🏼

PR #389, opened 2025-04-09
Logging refactoring: use `CanLog` directly in contexts

soareschen:
Follow up on: https://github.com/informalsystems/hermes-sdk/pull/590


ljoss17 (approved):
Looks good to me!

PR #392, opened 2025-04-09
fix: validator ID must be derived from its pub_key

### LINKED ISSUE #391: malicious validator can spoof votes from other validators
our cometbft verifier implementation is vulnerable to this recent bug at tendermint-rs

https://github.com/informalsystems/tendermint-rs/security/advisories/GHSA-6jrf-4jv4-r9mw

rnbguy:
closes #391

### DIFF
--- a/cairo-libs/packages/cometbft/src/types.cairo
+++ b/cairo-libs/packages/cometbft/src/types.cairo
@@ -453,7 +453,7 @@ pub struct Validator {
 #[generate_trait]
 pub impl ValidatorImpl of ValidatorTrait {
     fn validate_id(self: @Validator) {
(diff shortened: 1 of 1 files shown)


rnbguy:
actually, our code is not vulnerable to this bug. becuase we didn't incorporate [the `vote.verified` flag](https://github.com/informalsystems/tendermint-rs/blob/71c946084eceed3a2da91b78921d0345b46401e5/light-client-verifier/src/operations/voting_power.rs#L339) yet 😅 which lets the verifier skip the validation check for a vote if it's already validated -- even if it may be a different validator with a different public key, faking its account ID without actually signing the block.

soareschen (approved):

PR #393, opened 2025-04-10
Bootstrap both Anvil and Pathfinder in `MadaraBootstrap`

soareschen:


relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_anvil.rs
```diff
@@ -0,0 +1,71 @@
...
+            .map_err(Bootstrap::raise_error)?;
+
+        // Use RPC Port + 2 for Anvil port for now
+        let rpc_port = chain_node_config.rpc_port + 2;
+
+        let args = [
+            "--block-time",
+            "1",
+            "--chain-id",
+            "11155111",
```
rnbguy, line 46: is this value something significant?

rnbguy, line 46: is this value something significant?

relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_madara.rs
```diff
@@ -5,50 +5,71 @@ use hermes_cosmos_test_components::bootstrap::traits::chain::start_chain::{
...
             "--base-path",
-            &Runtime::file_path_to_string(chain_home_dir),
+            &Runtime::file_path_to_string(&madara_home),
+            "--rpc-port",
+            &rpc_port.to_string(),
+            "--gateway-port",
+            &gateway_port.to_string(),
             "--chain-config-override",
-            "block_time=1s,pending_block_update_time=1s",
+            "block_time=1s,pending_block_update_time=1s,chain_id=SN_SEPOLIA",
```
rnbguy, line 53: you pass `chain_id` to be `SN_SEPOLIA` here ..

relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_pathfinder.rs
```diff
@@ -0,0 +1,92 @@
...
+            "--storage.state-tries",
+            "archive",
+            "--ethereum.url",
+            &format!("http://localhost:{anvil_port}"),
+            "--gateway-url",
+            &format!("http://localhost:{gateway_port}/gateway"),
+            "--feeder-gateway-url",
+            &format!("http://localhost:{gateway_port}/feeder_gateway"),
+            "--chain-id",
+            "starknet-devnet",
```
rnbguy, line 63: .. but here, it is `starknet-devnet`.

rnbguy (approved):
LGTM 👍🏼

I added some comments.

relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_anvil.rs
soareschen: I think it is something that matches the Ethereum testnet ID. Otherwise Pathfinder would complain.

relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_madara.rs
soareschen: We also need this here, because Pathfinder would only work if it is connected to the Sepolia testnet. The main thing we need to be careful is that the signed transactions can also work on the testnet. But it is fine here because we are using test wallets generated by Madara.

relayer/crates/starknet-test-components/src/impls/bootstrap_madara/start_pathfinder.rs
soareschen: Yeah I'm not quite sure why we can set the chain ID here. Ideally it should be customizable on all 3 places, but right now it is only customizable here. It also don't appear to have any effect, since it is acting as a full node.

Issue #58, opened 2024-08-20
Implement pruning logic for old Tendermint consensus state in Cairo


Farhad-Shabani:
Implement pruning logic [here](https://github.com/informalsystems/ibc-starknet/blob/164be29d469171e955253823e2a306f52d3cc750/cairo-contracts/src/clients/tendermint/component.cairo#L216) for old Tendermint consensus states in order to prevent bloating states.

PR #394, opened 2025-04-11
feat: add consensus pruning logic

ljoss17:
Closes: #58 
Closes: #395

### Description

This PR adds pruning logic when updating clients and when recovering clients.

This PR also fixes client sequence incrementation to only increment when creating a new client


cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -346,11 +346,42 @@ pub mod CometClientComponent {
...
+                        .read_consensus_state(client_sequence, height.clone());
+
+                    let expiry_timestamp = consensus_state.timestamp.as_secs()
+                        + client_state.trusting_period.seconds;
+
+                    let host_timestamp = get_block_timestamp();
+
+                    if expiry_timestamp <= host_timestamp {
+                        let consensus_zero = CometConsensusStateImpl::zero();
+                        self.write_consensus_state(client_sequence, height.clone(), consensus_zero);
```
ljoss17, line 371: I didn't find a way to delete a Starknet storage `Map` entry.
Should the `Map` be rebuilt without the entry? @rnbguy

rnbguy: From https://book.cairo-lang.org/ch03-02-dictionaries.html#basic-use-of-dictionaries

> Before heading on and explaining how dictionaries are implemented it is worth mentioning that once you instantiate a Felt252Dict<T>, behind the scenes all keys have their associated values initialized as zero. This means that if for example, you tried to get the balance of an inexistent user you will get 0 instead of an error or an undefined value. This also means there is no way to delete data from a dictionary. Something to take into account when incorporating this structure into your code.

I guess we can't delete a data. We just set it to `Zero::zero` value.

cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
```diff
@@ -13,6 +13,10 @@ pub struct CometConsensusState {
 
 #[generate_trait]
 pub impl CometConsensusStateImpl of CometConsensusStateTrait {
+    fn zero() -> CometConsensusState {
+        CometConsensusState { timestamp: TimestampZero::zero(), root: StateRootZero::zero() }
+    }
```
rnbguy, line 18: you should rather implement `core::num::traits::Zero` for `CometConsensusState`.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -455,8 +486,18 @@ pub mod CometClientComponent {
...
+
+            let mut update_heights_span = update_heights.span();
+            let last_height = update_heights_span.pop_back().unwrap();
+
+            while let Option::Some(height) = update_heights_span.pop_front() {
+                let consensus_zero = CometConsensusStateImpl::zero();
+                self.write_consensus_state(subject_client_sequence, height.clone(), consensus_zero);
+            }
+
+            self.update_heights.write(subject_client_sequence, array![last_height.clone()]);
```
rnbguy, line 500: why do you keep the `last_height`? After the full client recovery, there should be only one consensus state -- the substitute consensus state.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -415,11 +415,42 @@ pub mod CometClientComponent {
...
+                if check_in_progress {
+                    let consensus_state = self
+                        .read_consensus_state(client_sequence, height.clone());
+
+                    let expiry_timestamp = consensus_state.timestamp.as_secs()
+                        + client_state.trusting_period.seconds;
+
+                    let host_timestamp = get_block_timestamp();
+
+                    if expiry_timestamp <= host_timestamp {
```
rnbguy, line 438: you can use the prexisting `consensus_state.state(client_state.trusting_period, Duration::zero())` to check if the consensus state is expired.

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -415,11 +415,42 @@ pub mod CometClientComponent {
...
+                        let consensus_zero = CometConsensusStateImpl::zero();
+                        self.write_consensus_state(client_sequence, height.clone(), consensus_zero);
```
rnbguy, line 440: as I suggested at other place, use a dedicated `remove_consensus_state` method for this.

cairo-contracts/packages/clients/src/cometbft/consensus_state.cairo
ljoss17: Done in 464b816bf4b4231e9e08f27ba015a1b8455e470d

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -455,8 +486,18 @@ pub mod CometClientComponent {
...
-            // They are untrusted now even if they are within trusted period.
+            let update_heights = self.read_update_heights(subject_client_sequence);
+            assert(update_heights.len() > 0, CometErrors::ZERO_UPDATE_HEIGHTS);
+
+            let mut update_heights_span = update_heights.span();
+            let last_height = update_heights_span.pop_back().unwrap();
+
+            while let Option::Some(height) = update_heights_span.pop_front() {
+                let consensus_zero = CometConsensusStateImpl::zero();
+                self.write_consensus_state(subject_client_sequence, height.clone(), consensus_zero);
```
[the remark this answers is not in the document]
ljoss17: Done in 3566f163c922482b3cbd1e79ebaf794805aa6cf8

cairo-contracts/packages/clients/src/cometbft/component.cairo
ljoss17: Done in cb64954b61ade3a79def6d6a03f31bc760cdadb6

cairo-contracts/packages/clients/src/cometbft/component.cairo
ljoss17: Done in 3566f163c922482b3cbd1e79ebaf794805aa6cf8

cairo-contracts/packages/clients/src/cometbft/component.cairo
```diff
@@ -455,8 +482,17 @@ pub mod CometClientComponent {
...
+            assert(update_heights.len() > 0, CometErrors::ZERO_UPDATE_HEIGHTS);
+
+            let mut update_heights_span = update_heights.span();
+            let last_height = update_heights_span.pop_back().unwrap();
+
+            while let Option::Some(height) = update_heights_span.pop_front() {
+                self.remove_consensus_state(subject_client_sequence, height.clone());
+            }
+
+            self.update_heights.write(subject_client_sequence, array![last_height.clone()]);
```
rnbguy, line 495:
```suggestion
            while let Option::Some(height) = update_heights_span.pop_front() {
                self.remove_consensus_state(subject_client_sequence, height.clone());
            }

            self.update_heights.write(subject_client_sequence, array![]);
```

cairo-contracts/packages/contracts/src/tests/client.cairo
```diff
@@ -207,3 +208,281 @@ fn test_client_recover_active_client() {
...
+    let _recover_resp = core.recover_client(msg.clone());
+
+    // -----------------------------------------------------------
+    // Check Results
+    // -----------------------------------------------------------
+
+    assert!(comet.status(0).is_active());
+
+    let third_consensus_state = comet.consensus_state_root(0, third_updating_height.clone());
+    assert!(third_consensus_state.is_non_zero());
```
rnbguy, line 485: why this should be non_zero? this should be pruned too.

ljoss17: Oh this was from before, when we were keeping latest height, will correct

rnbguy (approved):
✨

PR #424, opened 2025-05-07
feat: split ed25519 logic from Comet light client contract 

### LINKED ISSUE #422: split header verifier library from Comet contract
Currently, comet light client imports `cometbft` verifier logics -- which are just static functions. To reduce the size of comet contract code, we may deploy a static verifier class on Starknet and use [library dispatchers](https://book.cairo-lang.org/ch102-03-executing-code-from-another-class.html#library-calls) to use them.

- deploy the cometbft header verifier logic as a static contract.
- the comet light contract would use the library dispatcher to the above verifier class.

rnbguy:
closes #422


rnbguy:
not needed anymore because of #434

PR #426, opened 2025-05-13
ci: output logs even when tests succeed

ljoss17:
This PR adds the flag `--success-output=final` to `starknet-integration-tests` and `madara-integration-tests` so that logs are displayed even when tests succeeds in CI


rnbguy (approved):
👍🏼

PR #430, opened 2025-05-26
refactor: use `Array<u8>` over `ByteArray` and unify util functions

### LINKED ISSUE #407: unify cairo utils methods in a package
we are using a lot of `u8`, `u32`, `felt252`, `Array<u8>`, `[u32; 8]`, `ByteArray` etc. conversion from ics23 package. we should move all these generic cairo util methods to a separate package.

- byte/u32/felt conversions
- array/bytearray conversions
- `starkent::Store` implementations

### LINKED ISSUE #431: prefer `Array<u8>` over `ByteArray`
`ByteArray` is efficient for storage. But `Array<u8>` is efficient for runtime -- looping, querying bytes at an index.

Except for certain storage/serialization specific scenarios, we should use `Array<u8>` -- specifically for `protobuf` cairo package.

rnbguy:
closes #407
closes #431

- uses `Array<u8>` over `ByteArray` whenever possible
- unifies the spread out duplicated util functions under `ibc_utils` package


cairo-libs/packages/ibc_utils/Scarb.toml
```diff
@@ -0,0 +1,27 @@
...
+version       = { workspace = true }
+edition       = { workspace = true }
+cairo-version = { workspace = true }
+scarb-version = { workspace = true }
+authors       = { workspace = true }
+repository    = { workspace = true }
+license       = { workspace = true }
+Readme        = { workspace = true }
+keywords      = { workspace = true }
+description   = "Util methods for IBC"
```
ljoss17, line 12:
```suggestion
description   = "Utility methods for IBC"
```

ljoss17 (approved):
Looks good to me!
Very small nit, feel free to ignore if you prefer the current description

Issue #408, opened 2025-04-28
update `starknet-devnet` to `0.4.1`


rnbguy:
- bump to starknet rpc `0.8.1`.
- `starknet-devnet` has bumped execution gas amount to [`5_000_000_000`](https://github.com/0xSpaceShard/starknet-devnet/blob/aa338dc24d264fde463b9ac77af32b48802f95ed/crates/starknet-devnet-core/src/utils.rs#L56).
  - possibly removes the need of [our fork](https://github.com/0xSpaceShard/starknet-devnet/compare/main...informalsystems:starknet-devnet:rano/higher-sierra-gas).
- `starknet-devent` has removed `--request-body-size-limit` config argument in favor of [static values from mainnet](https://github.com/0xSpaceShard/starknet-devnet/blob/aa338dc24d264fde463b9ac77af32b48802f95ed/crates/starknet-devnet-core/src/constants.rs#L111-L113).

```rs
pub const MAXIMUM_CONTRACT_CLASS_SIZE: usize = 4_089_446;
pub const MAXIMUM_CONTRACT_BYTECODE_SIZE: usize = 81_920;
pub const MAXIMUM_SIERRA_LENGTH: usize = 81_920;
```

- They are checked when [a class is declared](https://github.com/0xSpaceShard/starknet-devnet/blob/aa338dc24d264fde463b9ac77af32b48802f95ed/crates/starknet-devnet-core/src/starknet/add_declare_transaction.rs#L36).
- But this causes IBC Core and Comet client contract classes to hit the contract bytecode limit.
  - IBCCore: `Declaring class: serialized size: 1937183 bytes, sierra: 37923 felts, casm: 85663 felts`
  - CometClient: `Declaring class: serialized size: 4089220 bytes, sierra: 64599 felts, casm: 172534 felts`

rnbguy:
We have this code in our repo.

https://github.com/informalsystems/ibc-starknet/blob/d69433c7fc45bdac8acf83a5bd878750b835b4dc/relayer/crates/starknet-chain-components/src/impls/contract/declare.rs#L68-L70

Probably `180000` is too high value and not ideal for Starknet mainnet/testnet.

rnbguy:
Looks like even `starkli` can't compile the sierra files of `IBCCore` and `Comet` contract. It throws

```
Error: CompilationFailed
```

rnbguy:
The first commit to introduce this limit overflow is 46e9240 (via `git bisect`)

[...]

PR #434, opened 2025-05-29
fix: library calls to reduce casm size

### LINKED ISSUE #422: split header verifier library from Comet contract

### LINKED ISSUE #423: split ics23 verifier library from Comet contract
Currently, comet light client imports `ics23` cosmos commitment vector verifier logics -- which are just static functions. To reduce the size of comet contract code, we may deploy a static verifier class on Starknet and use [library dispatchers](https://book.cairo-lang.org/ch102-03-executing-code-from-another-class.html#library-calls) to use them.

- deploy the ics23 commitment vector verifier logic as a static contract.
- the comet light contract would use the library dispatcher to the above verifier class.

rnbguy:
Closes #422
Closes #423

This PR introduces three library contracts:

- cometbft: verify header and misbehavior
- ics23: verify membership and non-membership
- protobuf:
  - encode: client state, channel end, connection end
  - decode: comet header, merkle proof

All these calls are static (i.e. no storage access required).

CometClient and IBCCore contract uses these library contracts via `LibraryDispatcher`. This way we can avoid importing these heavy libraries inside these two contracts -- resulting a reduced casm size.

With this PR, we move back to original `starknet-devnet` releases without requiring to patch size limits (#408).

- [x] fix cairo integration tests
- [x] avoid hard coding library contract classes
- [x] rebase on `main` after #430 is merged


rnbguy:
the contracts are deployable on testnet again !
https://sepolia.voyager.online/contract/0x013b8f63d5baf8cd7eaea69968e0b68df4450196183a568395744291006e59f6

ljoss17 (approved):
Looks good to me, left a small question in github workflow file

PR #442, opened 2025-06-24
feat: log gas usage of starknet transactions

### LINKED ISSUE #441: benchmark gas costs of cairo contracts
We are trying to analyze the gas costs of our contracts -- if they are too big for `mainnet`.

For this, we need to trace the gas usage of the transactions in our e2e tests.

The [limit on sierra gas](https://docs.starknet.io/resources/chain-info/#current_limits) is per tx is 1 million and per block is 5 million.

rnbguy:
closes #441

it includes few chores regarding dependencies 

- upgrade snfoundry, openzeppelin
- pin cgp to fixed commit
- upgrade rust dependencies

### DIFF
--- a/relayer/crates/starknet-chain-components/src/impls/tx_response.rs
+++ b/relayer/crates/starknet-chain-components/src/impls/tx_response.rs
@@ -1,7 +1,7 @@
 use core::time::Duration;
 
 use hermes_core::logging_components::traits::CanLog;
-use hermes_core::logging_components::types::LevelInfo;
+use hermes_core::logging_components::types::LevelDebug;
 use hermes_core::relayer_components::transaction::traits::{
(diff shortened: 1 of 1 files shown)


relayer/crates/starknet-chain-components/src/impls/tx_response.rs
```diff
@@ -41,7 +44,20 @@ where
...
+                chain
+                    .log(
+                        &format!(
+                            "[{}] {:?}",
+                            tx_hash.to_fixed_hex_string(),
+                            tx_response.execution_resources(),
+                        ),
+                        &LevelInfo,
+                    )
+                    .await;
```
ljoss17, line 58: This is to output the gas usage of a transaction for benchmarking right? If so it should be at debug level as this doesn't feel very useful for general purpose relaying from an operator's point of view.

ljoss17 (approved):
Pre-approving, but left a comment/question

PR #453, opened 2025-07-25
Run client refresh test

ljoss17:
This PR adds client refresh test for Starknet to Cosmos relaying.

Client refresh test is only running for Cosmos client since the Starknet client can't expire.

### DIFF
--- a/relayer/crates/starknet-cli/src/commands/start.rs
+++ b/relayer/crates/starknet-cli/src/commands/start.rs
@@ -44,9 +44,9 @@ pub struct StartRelayerArgs {
     #[clap(long = "stop-after-blocks", required = false)]
     stop_after_blocks: Option<humantime::Duration>,
 
-    #[clap(long = "refresh-rate-a-to-b", required = false)]
+    #[clap(long = "refresh-rate-a", required = false)]
     refresh_rate_a: Option<humantime::Duration>,
 
-    #[clap(long = "refresh-rate-b-to-a", required = false)]
+    #[clap(long = "refresh-rate-b", required = false)]
     refresh_rate_b: Option<humantime::Duration>,
 }


relayer/crates/starknet-integration-tests/src/contexts/setup.rs
```diff
@@ -60,6 +63,18 @@ impl StarknetTestSetup {
         starknet_builder: StarknetBuilder,
         wasm_code_hash: [u8; 32],
     ) -> Self {
+        let cosmos_create_client_payload_options =
+            if let Ok(raw_trusting_period) = var("NEW_TRUSTING_PERIOD") {
```
rnbguy, line 67:
```suggestion
            if let Ok(raw_trusting_period) = var("COSMOS_TRUSTING_PERIOD") {
```

relayer/crates/starknet-cli/src/commands/start.rs
```diff
@@ -43,4 +43,10 @@ pub struct StartRelayerArgs {
 
     #[clap(long = "stop-after-blocks", required = false)]
     stop_after_blocks: Option<humantime::Duration>,
+
+    #[clap(long = "refresh-rate-a-to-b", required = false)]
+    refresh_rate_a: Option<humantime::Duration>,
+
+    #[clap(long = "refresh-rate-b-to-a", required = false)]
+    refresh_rate_b: Option<humantime::Duration>,
```
rnbguy, line 51:
```suggestion
    #[clap(long = "refresh-rate-a", required = false)]
    refresh_rate_a: Option<humantime::Duration>,

    #[clap(long = "refresh-rate-b", required = false)]
    refresh_rate_b: Option<humantime::Duration>,
```

actually do we even need the `long` attribute? it's the same as the field names.

.github/workflows/hermes.yaml
```diff
@@ -119,7 +119,115 @@ jobs:
               --workspace \
               --retries 2 \
               --test-threads=$TEST_THREADS \
-              --success-output=final
+              --success-output=final \
+              --filter-expr '!(test(modified_test_suite::))'
```
rnbguy, line 123: do we really need a separate workflow test for this ?

you can just add next test here with new environment variables.

```
NEW_TRUSTING_PERIOD=30 COSMOS_REFRESH_RATE=20 \
          nix shell \
            ..#osmosis \
            ..#wasm-simapp \
            ..#rust \
            ..#cargo-nextest \
            -c \
            cargo nextest run --release \
              --workspace \
              --retries 2 \
              --test-threads=$TEST_THREADS \
              --success-output=final \
              -- modified_test_suite

```

rnbguy: actually, having separate runs for unit tests is not a good practice. we need to tackle this cleanly in separate issue.

rnbguy (approved):
approved but left some comments.

.github/workflows/hermes.yaml
ljoss17: I will open an issue then.
The ideal solution would be an `OverwriteConfig` component to allow individual tests to overwrite the necessary configurations, similar to what we had in Hermes v1

ljoss17: https://github.com/informalsystems/hermes-sdk/issues/626

rnbguy: for setting specific environment variables for specific tests, we can use `assay` crate.

ref: https://github.com/mgattozzi/assay?tab=readme-ov-file#env-vars

Issue #374, opened 2025-03-21
permissionless starknet wasm light client


rnbguy:
after we have resolved #370 (starknet header validation) and #371 (starknet storage proof validation), we will finish the wasm light client with proper algorithm to achieve permissionless relaying.

rnbguy:
We can test this using different relayer wallets on:

- local devnets (write an e2e test)
- public testnets (perform manually)

ps. for the public testnets, we are currently blocked. we can't store our wasm starknet light client on Osmosis because of tx gas limits.

PR #454, opened 2025-07-29
test: permisionless relaying

ljoss17:
Closes: #374 

~Blocked by: https://github.com/informalsystems/hermes-sdk/pull/621 & https://github.com/informalsystems/hermes-sdk/pull/623~ Resolved

This PR sets the tests to run with different wallets to test permissionless relaying


relayer/crates/starknet-integration-tests/src/impls/build_bootstrap.rs
```diff
@@ -120,7 +125,7 @@ where
 
         let proof_signer = Secp256k1KeyPair::from_mnemonic(
             bip39::Mnemonic::from_entropy(
-                &relayer_wallet.signing_key.to_bytes_be(),
+                &relayer_wallet_1.signing_key.to_bytes_be(),
```
rnbguy, line 128: what was `proof_signer` again ?

I think, this was part of the permissioned relaying.

relayer/crates/starknet-integration-tests/src/tests/erc20.rs
```diff
@@ -141,7 +186,6 @@ fn test_erc20_transfer() -> Result<(), Error> {
 
             match &erc20_events[0] {
                 Erc20Event::Transfer(transfer) => {
-                    assert_eq!(transfer.from, account_address);
```
rnbguy, line 144: why this line is deleted ?

relayer/crates/starknet-integration-tests/src/tests/erc20.rs
```diff
@@ -162,10 +206,6 @@ fn test_erc20_transfer() -> Result<(), Error> {
 
             info!("recipient balance transfer: {}", recipient_balance_b);
 
-            assert_eq!(
-                sender_balance_b.quantity,
-                sender_balance_a.quantity - transfer_amount
-            );
```
rnbguy, line 168: why is this assert removed ?

rnbguy (commented):
left some comments.

relayer/crates/starknet-integration-tests/src/tests/erc20.rs
ljoss17: We were not sending tokens from user 1 to user 2, but from relayer wallet to another address.
Since we now use multiple relayer wallets this assertion isn't correct anymore

relayer/crates/starknet-integration-tests/src/impls/build_bootstrap.rs
ljoss17: Oh good catch, it seems we don't use the proof signer anymore. I will remove it.

ljoss17: Done here 0dde7f5f2b3bcda70e370504eb685c25047936d5

relayer/crates/starknet-integration-tests/src/tests/erc20.rs
rnbguy: hmm. can we use `send_messaged_with_signer` to keep these asserts ?

PR #459, opened 2025-08-05
ci: update madara binary

rnbguy:
build madara from `main` branch

### DIFF
--- a/.github/workflows/hermes.yaml
+++ b/.github/workflows/hermes.yaml
@@ -142,7 +142,7 @@ jobs:
         with:
           repository: madara-alliance/madara
           path: madara
-          ref: e6fca12103a97238b5b5a87b5a22342804caac4c # chore/bump-blockifier
+          ref: 67a2ac59af4c2d98216b8fb2ef1e538020cbf462 # main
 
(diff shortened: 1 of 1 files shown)


.github/workflows/hermes.yaml
```diff
@@ -22,22 +22,26 @@ jobs:
...
       matrix:
-        os: [ubuntu-22.04, ubuntu-22.04-arm, macos-14]
+        os: [ubuntu-22.04, ubuntu-22.04-arm, macos-13]
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    timeout-minutes: 120
     steps:
       - name: Free Disk Space (Ubuntu)
-        if: ${{ matrix.os != 'macos-14' }}
+        if: ${{ matrix.os != 'macos-13' }}
```
rnbguy, line 30: note that: `macos-14` is arm64 but `macos-13` is amd64.

we need this as the latest madara build steps use docker. `macos-14` (arm64) and `macos-15` (arm64) do not support docker because they don't support nested virtualization.

ljoss17 (approved):
Looks good to me!

